PR #87, opened 2014-08-15
Rename to be more descriptive and take a project object.

bwendling:
The 'relative' function was only being used to get the path relative to the project's root. This changes the API to take the project object and renamed to make it explicit what's happening.

### DIFF
--- a/docs/library.rst
+++ b/docs/library.rst
@@ -310,6 +310,17 @@ resource.
   module_name = libutils.modname(resource)
 
 
+``libutils.path_relative_to_project_root()``
+--------------------------------------------
+
+Retrieve the path relative to the project's root directory.
+
+.. code-block:: python
+
+  # Get the path relative to the project's root directory.
+  relpath = libutils.relative(myproject.address, path)
+
+
 ``libutils.path_to_resource()``
 -------------------------------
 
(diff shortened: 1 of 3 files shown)


CAM-Gerlach:
Sorry to revive this old thread, but as one of the developers of the Spyder IDE I've run into a bevy of false ``DeprecationWarnings`` due to this due to this patch, despite us only ever using the undeprecated ``path_to_resource`` function. Due to the design of this patch, many unit testing frameworks as well as any setting of the warning filter to ``"default"`` (e.g. ``-Wd`` option in python, pytest, etc), as is common when checking for deprecation and other warnings in test suites (as I was) will trigger false ``DeprecationWarnings`` to be reported.

This is, of course, due to calling the deprecated function from its "modern" replacement, with the latter as a wrapper of the former, rather than the other way around or both calling a common lower-level function. Therefore, would you be willing to refactor this to behave properly by your preferred means (common low-level function, wrapping the other way around, etc), or willing to review a PR doing the same (by whichever means you suggest)? Thanks.

mcepl:
@CAM-Gerlach Thanks for the notice. Of course, pull request would be warmly considered.

CAM-Gerlach:
Okay, thanks @mcepl ! Will do, hopefully tomorrow.

Issue #102, opened 2015-01-08
Leaked variable from list comprehension in method extraction.


Ferada:
I've come across some unexpected behaviour while extracting a method, although I'm not sure if it's not still technically correct.

A reduce example is something like the following:

    def foo():
        x = [e for e in []]

        for e in []:
            def bar():
                e[42] = 1

Marking the complete loop body and running method extraction will produce the following result:

    def foo():
        x = [e for e in []]

        baz(e)

    def baz(e):
        for e in []:
            def bar():
                e[42] = 1

What is unexpected is the `e` parameter for `baz`.  If the `x = ...` is removed, that parameter isn't added, so it looks to me like it's including the leaking variable from the list comprehension (something
you can also stumble upon in general), although the `e` from the `for`-loop will definitely supersede it.

mcepl:
@aligrudi what do you think? It truly looks like a bug to me.

aligrudi:
Matěj Cepl <notifications@github.com> wrote:
> @aligrudi what do you think? It truly looks like a bug to me.

Indeed.  To fix it, the _For() method of _FunctionInformationCollector
in rope/refactor/extract.py has to be modified.  I wonder why
node.target does not have a Name node with ast.Store context.

	Ali

PR #103, opened 2015-01-11
Fix extract method with list comprehension.

Ferada:
This should fix https://github.com/python-rope/rope/issues/102. I've added three test cases which test my original problem and then also two variants on that where the extraction should actually add another read parameter and one where the `else` case of the loop reads variables.

Now I've implemented this by looking at that file, but really, if you can suggest anything to clean it up I'd be very grateful.

### DIFF
--- a/rope/refactor/extract.py
+++ b/rope/refactor/extract.py
@@ -610,7 +610,8 @@ class _FunctionInformationCollector(object):
     def _read_variable(self, name, lineno):
         if self.start <= lineno <= self.end:
             if name not in self.written:
-                self.read.add(name)
+                if not self.conditional or name not in self.maybe_written:
+                    self.read.add(name)
         if self.end < lineno:
             if name not in self.postwritten:
                 self.postread.add(name)
@@ -670,7 +671,18 @@ class _FunctionInformationCollector(object):
         self._handle_conditional_node(node)
 
     def _For(self, node):
-        self._handle_conditional_node(node)
+        self.conditional = True
+        try:
+            # iter has to be checked before the target variables
+            ast.walk(node.iter, self)
+            ast.walk(node.target, self)
+
+            for child in node.body:
+                ast.walk(child, self)
+            for child in node.orelse:
+                ast.walk(child, self)
+        finally:
+            self.conditional = False
 
 
 def _get_argnames(arguments):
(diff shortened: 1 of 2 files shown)


aligrudi:
Olof-Joachim Frahm <notifications@github.com> wrote:
> This should fix https://github.com/python-rope/rope/issues/102. I've added three test cases which test my original problem and then also two variants on that where the extraction should actually add another read parameter and one where the `else` case of the loop reads variables.
> 
> Now I've implemented this by looking at that file, but really, if you can suggest anything to clean it up I'd be very grateful.

Looks good to me.

However I wonder if this idea works too: when in a conditional
block (including for loops), why not append the variable
written to to self.written and only when the loop finishes
moving them to self.maybe_written or maybe something like
this (not tested and just an idea that may not work):

diff --git a/rope/refactor/extract.py b/rope/refactor/extract.py
--- a/rope/refactor/extract.py
+++ b/rope/refactor/extract.py
@@ -610,7 +610,8 @@ class _FunctionInformationCollector(object):
     def _read_variable(self, name, lineno):
         if self.start <= lineno <= self.end:
             if name not in self.written:
-                self.read.add(name)
+                if not self.conditional or name not in self.maybe_written:
+                    self.read.add(name)
         if self.end < lineno:
             if name not in self.postwritten:
                 self.postread.add(name)


(Handling nested conditional blocks is more tricky, I guess).

I also wonder if the current implementation handles the
following cases as well:

  for x in []:
      x = 1		# extract begin
      y = x		# extract end

Here x should not be passed to the extracted function.  And:

  x = 1
  for x in []:		# extract begin
  	x = 1
  y = x			# extract end

Here x should be passed to the extracted function.

	Ali

Ferada:
> I also wonder if the current implementation handles the following
> cases as well:

As far as I can see the master branch mishandles the first case, but handles the second case fine.  Both with your suggestion above, or my pull request the first case is correctly handled.

> However I wonder if this idea works too:

I've tried your change and while I still had to modify `_For` to change the walking order (`iter`, what the loop is iterating over, first, then `target`), it works, so I've uploaded that version instead.  Thanks for the help, that way it's absolutely easier!

aligrudi:
Olof-Joachim Frahm <notifications@github.com> wrote:
> > I also wonder if the current implementation handles the following
> > cases as well:
> 
> As far as I can see the master branch mishandles the first case, but handles the second case fine.  Both with your suggestion above, or my pull request the first case is correctly handled.
> 
> > However I wonder if this idea works too:
> 
> I've tried your change and while I still had to modify `_For` to change the walking order (`iter`, what the loop is iterating over, first, then `target`), it works, so I've uploaded that version instead.  Thanks for the help, that way it's absolutely easier!

ACK

Thanks,
Ali

Issue #106, opened 2015-02-10
failing tests on Windows


mcepl:
See https://gitlab.com/mcepl/rope/snippets/3192 for the full log.

Some issues are clearly Windows weirdness (e.g., subprocess), but some are looking as rope lacks some stability of the code (e.g., `<rope.base.pyobjectsdef.PyClass object at 0x02981870> != <rope.base.pyobjectsdef.PyClass object at 0x029B4930>`)

aligrudi:
> See https://gitlab.com/mcepl/rope/snippets/3192 for the full log.
> 
> Some issues are clearly Windows weirdness (e.g., subprocess), but some are looking as rope lacks some stability of the code (e.g., `<rope.base.pyobjectsdef.PyClass object at 0x02981870> != <rope.base.pyobjectsdef.PyClass object at 0x029B4930>`)

The main problem seems to be in DynamicOITest; rope needs to execute
python modules in a sub-process to collect the type of objects passed
to functions.  This seems to fail: either rope fails to start the
sub-process or it cannot receive the information the process sends
back.  Instead of fixing it, it may be acceptable to add a test
decorator (@onlyunix for instance) to disable them on Windows.

	Ali

mcepl:
I played with it for a moment and I am returning to the `run` method in `doa.py` module.

1. When running on Linux v. Windows, the args variable is very different. On Windows it is
```
['C:\\Python27\\python.exe', 'C:\\Users\\ceplm\\D...t\\mod.py']
```
(yes, I hope the string is actually complete, just Visual Studio is not willing to give me better representation), whereas on Linux it is
```
['/usr/bin/python',
 '/home/matej/archiv/knihovna/repos/rope/rope/base/oi/runmod.py',
 '3037',
 '/dev/shm/sample_project',
 '/dev/shm/sample_project/mod.py']
```
(yes, Eclipse provides a way more civilized output)
It seems like we run completely different Python script, and also all remaining parameters are missing. 
2. What is `executable` parameter of `subprocess.Popen` good for?
3. both `stdout` and `stderr` parameters point to the same file `sys.stdout`. Is it correct? Do we need `stderr` at all? Or shouldn't we collect their content to two variables and combine them later if necessary?

mcepl:
> Instead of fixing it, it may be acceptable to add a test decorator (@onlyunix for instance) to disable them on Windows.

[...]

PR #108, opened 2015-02-12
Prefer unittest skip decorators over our own.

mcepl:
When thinking about #106, it seems to me it would be cleaner to use `unittest` standard skip-decorators instead of our own.

@aligrudi , what do you think?

### DIFF
--- a/ropetest/codeanalyzetest.py
+++ b/ropetest/codeanalyzetest.py
@@ -432,7 +432,7 @@ class ScopeNameFinderTest(unittest.TestCase):
         found_pyname = name_finder.get_pyname_at(code.index('afunc') + 1)
         self.assertEquals(afunc.get_object(), found_pyname.get_object())
 
-    @testutils.run_only_for_25
+    @testutils.only_for('2.5')
     def test_relative_modules_after_from_statements(self):
         pkg1 = testutils.create_package(self.project, 'pkg1')
         pkg2 = testutils.create_package(self.project, 'pkg2', pkg1)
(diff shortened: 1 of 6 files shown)


mcepl:
@aligrudi  ping?

aligrudi:
> @aligrudi  ping?

Seems OK.

Thanks,
Ali

ropetest/codeanalyzetest.py
```diff
@@ -432,7 +432,7 @@ def test_renaming_functions_with_from_import_and_parens(self):
         found_pyname = name_finder.get_pyname_at(code.index('afunc') + 1)
         self.assertEquals(afunc.get_object(), found_pyname.get_object())
 
-    @testutils.run_only_for_25
```
hayd: is 2.5 supported? If not, perhaps these tests should be removed? :s

mcepl: > is 2.5 supported? If not, perhaps these tests should be removed? :s

From time to time (not very often, I am afraid) I pull out a virtual machine with RHEL-5 (that's python 2.4) and try to run the tests. Last time I tried (and that is some time, true) all tests passed. But perhaps, it is really the time. @aligrudi , what do you think, should we kill python < 2.6?

hayd: (It may no longer be the case due to my py3 PR, I think the new exception syntax was backported as far back as py26 only... I did not realise there was 2.5- support. :s )

mcepl: > (It may no longer be the case due to my py3 PR, I think the new exception syntax was backported as far back as py26 only... I did not realise there was 2.5- support. :s )

Right. So we lost < 2.6 compatibility. Not a big deal. I don't have to look for that VM ;).

Issue #109, opened 2015-03-05
Consider fixes from python-mode


mcepl:
Not that it make much sense to me, but we have
- [ ] https://github.com/klen/python-mode/commit/cec42722bf3f217da1ffbaa5c0486e3dfb08a377
- [x] https://github.com/klen/python-mode/commit/70d8eab3bee42fbde292b4fa2813e69d9a5e45fe

aligrudi:
Matěj Cepl <notifications@github.com> wrote:
> Not that it make much sense to me, but we have
>  * [ ] https://github.com/klen/python-mode/commit/cec42722bf3f217da1ffbaa5c0486e3dfb08a377

This seems OK.

>  * [ ] https://github.com/klen/python-mode/commit/70d8eab3bee42fbde292b4fa2813e69d9a5e45fe

I am not sure about this one:

> diff --git a/pylibs/rope/base/pycore.py b/pylibs/rope/base/pycore.py
> --- a/pylibs/rope/base/pycore.py
> +++ b/pylibs/rope/base/pycore.py
> @@ -306,7 +305,7 @@ def __init__(self, pycore):
>  
>      def _invalidate_resource(self, resource):
>          if resource in self.module_map:
> -            self.forget_all_data()
> +            self.module_map[resource]._forget_concluded_data()
>              self.observer.remove_resource(resource)
>              del self.module_map[resource]

I think self.forget_all_data() cannot be removed: if "resource"
is imported in other modules, they need to be invalidated as well.

> @@ -258,7 +258,6 @@ def analyze_module(self, resource, should_analyze=lambda py: True,
>          if followed_calls is None:
>              followed_calls = self.project.prefs.get('soa_followed_calls', 0)
>          pymodule = self.resource_to_pyobject(resource)
> -        self.module_cache.forget_all_data()
>          rope.base.oi.soa.analyze_module(
>              self, pymodule, should_analyze, search_subscopes, followed_calls)

I am not sure about this one; I guess forget_all_data() is called
to invalidate cached PyObjects, before starting static object
analysis.

	Ali

mcepl:
> > Not that it make much sense to me, but we have * [ ] https://github.com/klen/python-mode/commit/cec42722bf3f217da1ffbaa5c0486e3dfb08a377
> This seems OK.

[...]

PR #110, opened 2015-03-08
Add a little explanation of what _realpath function is about.

mcepl:
While looking at #109, I was digging into what `_realpath` is about and I thought it to be worthy to preserve this explanation.

@aligrudi , what do you think?

### DIFF
--- a/rope/base/project.py
+++ b/rope/base/project.py
@@ -453,6 +453,11 @@ def _realpath(path):
 
     Is equivalent to ``realpath(abspath(expanduser(path)))``.
 
+    Of the particular notice is the hack dealing with the unfortunate
+    sitaution of running native-Windows python (os.name == 'nt') inside
+    of Cygwin (abspath starts with '/'), which apparently normal
+    os.path.realpath completely messes up.
+
     """
     # there is a bug in cygwin for os.path.abspath() for abs paths
     if sys.platform == 'cygwin':


aligrudi:
> While looking at #109, I was digging into what `_realpath` is about and I thought it to be worthy to preserve this explanation.
> 
> @aligrudi , what do you think?
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/python-rope/rope/pull/110

Unfortunately I do not remember the exact problem, but judging
by the comment preceding the 'cygwin' check, I agree that
it is a good idea to document it explicitly.

Thanks,
Ali

mcepl:
Alternative would be to do something like this:
```diff
--- Lib/site-packages/pyx/filelocator.py-orig	2012-10-24 10:55:17.604624800 +0200
+++ Lib/site-packages/pyx/filelocator.py	2012-10-24 10:55:22.845624800 +0200
@@ -201,6 +201,11 @@
         else:
             return []
         full_filename = full_filenames.split("\n")[0].rstrip("\r")
+
+        # Detect Cygwin kpsewhich on Windows Python
+        if os.name == 'nt' and full_filename.startswith('/'):
+            full_filename = pycompat.popen('cygpath -w "%s"' % full_filename).read().strip()
+
         def _opener():
             try:
                 return builtinopen(full_filename, mode)
```

aligrudi:
> Alternative would be to do something like this:
> ```diff
> --- Lib/site-packages/pyx/filelocator.py-orig	2012-10-24 10:55:17.604624800 +0200
> +++ Lib/site-packages/pyx/filelocator.py	2012-10-24 10:55:22.845624800 +0200
> @@ -201,6 +201,11 @@
>          else:
>              return []
>          full_filename = full_filenames.split("\n")[0].rstrip("\r")
> +
> +        # Detect Cygwin kpsewhich on Windows Python
> +        if os.name == 'nt' and full_filename.startswith('/'):
> +            full_filename = pycompat.popen('cygpath -w "%s"' % full_filename).read().strip()
> +

Since I do not know much about cygwin, I cannot make a
helpful comment.  However, executing cygpath on each
_realpath() call can be very expensive?

	Ali

mcepl:
Probably, and I don't have any machine with Cygwin around (and not that much interest). I was just more or less think aloud. Let’s it rest and wait if somebody complains.

PR #111, opened 2015-03-10
Wrong parameter default value in Inline Parameter

Sathors:
The default value in the Inline Parameter example was 1, and when using Inline was passing 2.

### DIFF
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -551,7 +551,7 @@ after inlining p2 parameter will have:
 
 .. code-block:: python
 
-  def f(p1=1, p2=1):
+  def f(p1=1, p2=2):
       pass
 
   f(3, 2)


mcepl:
@aligrudi does it make any sense to you? Do we have a bug there?

Sathors:
Well, trying the example in the doc will give you:
```python
def f(p1=1, p2=1):
    pass
 
    f(3, 1)
    f(p2=1)
    f(3, 4)
```
when the documentation says it should give
```python
    f(3, 2)
    f(p2=2)
    f(3, 4)
```

To obtain the expected result, we just have to change the default value of the P2 inlined parameter from 1 to 2.

This is not a bug, but a not-working example in the documentation.

mcepl:
> This is not a bug, but a not-working example in the documentation.

aka bug in the documentation, right. Thank you.

aligrudi:
> > This is not a bug, but a not-working example in the documentation.
> 
> aka bug in the documentation, right. Thank you.

Indeed.

Thanks,
Ali

PR #116, opened 2015-05-22
Patchedast improvements

enomado:
Hello. Its my second attempt to push all its stuff to main repo. 

This code was created while running over my projects, libraries and received parsing errors. I do not expect this pull request will be approved soon, but is can be helpful for some people.

I dont know how you guys didnt stuck on constructions like`"foo" "bar"` or `call(1, a=1, *[1]` so long.

Code is very dirty, but correct solution could be in patch python itself

### DIFF
--- a/rope/refactor/patchedast.py
+++ b/rope/refactor/patchedast.py
@@ -247,10 +247,7 @@ class _PatchingASTWalker(object):
     def _Assign(self, node):
         children = self._child_nodes(node.targets, '=')
         children.append('=')
-        if isinstance(node.value, ast.Str):
-            children.append(self.Not_multiline_string)
-        else:
-            children.append(node.value)
+        children.append(node.value)
         self._handle(node, children)
 
     def _AugAssign(self, node):
--- a/rope/refactor/patchedast.py
+++ b/rope/refactor/patchedast.py
@@ -73,6 +73,7 @@ class _PatchingASTWalker(object):
     Number = object()
     String = object()
     semicolon_or_as_in_except = object()
+    Not_multiline_string = object()
 
     def __call__(self, node):
         method = getattr(self, '_' + node.__class__.__name__, None)
@@ -110,7 +111,11 @@ class _PatchingASTWalker(object):
             else:
                 if child is self.String:
                     region = self.source.consume_string(
-                        end=self._find_next_statement_start())
+                        end=self._find_next_statement_start(), not_multiline = False)
+                elif child is self.Not_multiline_string:
+                    region = self.source.consume_string(
+                        end=self._find_next_statement_start(), not_multiline = True)
+                                
                 elif child is self.Number:
                     region = self.source.consume_number()
                 elif child == '!=':
@@ -242,7 +247,10 @@ class _PatchingASTWalker(object):
     def _Assign(self, node):
         children = self._child_nodes(node.targets, '=')
         children.append('=')
-        children.append(node.value)
+        if isinstance(node.value, ast.Str):
+            children.append(self.Not_multiline_string)
+        else:
+            children.append(node.value)
         self._handle(node, children)
 
     def _AugAssign(self, node):
@@ -259,7 +267,14 @@ class _PatchingASTWalker(object):
         self._handle(node, children)
 
     def _BoolOp(self, node):
-        self._handle(node, self._child_nodes(node.values,
+        children = []
+        for no in node.values:
+            if isinstance(no, ast.Str):
+                children.append(self.Not_multiline_string)
+            else:
+                children.append(no)
+
+        self._handle(node, self._child_nodes(children,
                                              self._get_op(node.op)[0]))
 
     def _Break(self, node):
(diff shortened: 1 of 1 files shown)


enomado:
Almost works except sorted_children.
Some of them I can fix only by changing unit test expectations, because ast represents multiline strings by only one node.

enomado:
Another way is to make pypy parser create `region` field.

mcepl:
OK, two things:
1. test suite will be passing,
2. each commit will constitute single logical change with the commit message describing such change. E.g.  https://github.com/enomado/rope/commit/ff2cf59cd1ef2d491a07bfa614256a9981d7fc89 with the commit message "Clean" is not acceptable.
Only then we can start to go through the changes to decide what to do about them.

enomado:
Thanks. 

I have even better plan. Now I'm investigating pypy parser. I could adapt it to bring offset of the end of ast node. Here it is: https://bitbucket.org/pypy/pypy/src/default/pypy/interpreter/astcompiler/astbuilder.py
As you can see it is almost the same as rope's. There are lots of work and I really dont know what to do with that stuff - merge to pypy or fork it and update when pypy code change as well as I dont know how you guys will concern to merge pypy code to rope...

mcepl:
@aligrudi ???

I don’t think we will ever merge anything requiring so many dependencies as that part of PyPI. Rope is in more or less in the maintenance mode and pulling in that code of PyPI requires more or less fork of the current project. Feel free to do it if you wish, but it is a different project then.

Concerning your original pull request, let us take a look.

aligrudi:
> @aligrudi ???
> 
> I don’t think we will ever merge anything requiring so many dependencies as that part of PyPI. Rope is in more or less in the maintenance mode and pulling in that code of PyPI requires more or less fork of the current project. Feel free to do it if you wish, but it is a different project then.

Although relying on an internal parser inside rope itself
would be wonderful, I agree that it is very difficult to
maintain it.  Thus, our best choice is the native (and
fast) python built-in parser.

About the patch.  I had a quick look; a few comments:

It would be great if you include the python code that shows
the problem, ideally as a test case inside patchedasttest.py.

About the multi-line string problem, is it not possible
to update the regular expression pattern in consume_string()
to cover that case?

Thanks,
Ali

enomado:
Okay lets talk about problems...

1. DictComp, Set, SetComp nodes.
2. Some statistics:
https://gist.github.com/enomado/d90bc95a95dafce68c30
gives overall: 2342, failed: 56
And with my patches. No so good as I expected...
overall: 2342, failed: 40
3.

enomado:
Okay lets talk about problems...

DictComp, Set, SetComp nodes.
Some statistics: https://gist.github.com/enomado/d90bc95a95dafce68c30 
gives overall: 2342, failed: 56 

And with my patches. No so good as I expected... 
overall: 2342, failed: 40



> About the multi-line string problem, is it not possible
> to update the regular expression pattern in consume_string()
> to cover that case?

I think strings, list comprehensions, function args cant be parsed by regexp. If I was a computer scientist i'd said something like 'regular grammar is not applicable', but I'm not sure and just show you a few examples.

This is a single Str node
```python
a = (((('' '' '' # test test
	'' \
	''' '''))))
```

This is list comprehension in function args 
```python
fun_call(float(y) for y in x for x in l)
````

As I said, function call can contain almost anything in random order.
```python
call(1, a=1, *[1, "" ""], c = 1, wryyy=lambda r=lambda: x : {'b':1}, **{'b':1})
````
**This can only be parsed by fully functional python language parser. I believe everyone agree with that on this stage.**

And we have few options:
* Take cpython's parser, written in C, add needed info, and create cpython fork/library.
* Take pypy's parser, written in Python. I take pypy and I already know how to patch it.

aligrudi:
Stanislav <notifications@github.com> wrote:
> I think strings, list comprehensions, function args cant be parsed by regexp. If I was a computer scientist i'd said something like 'regular grammar is not applicable', but I'm not sure and just show you a few examples.
> 
> This is a single Str node
> ```python
> a = (((('' '' '' # test test
> 	'' \
> 	''' '''))))
> ```

Regular expressions may not be a appropriate tool for handling
problematic AST nodes, but for strings we already have:

    def consume_string(self, end=None):
        if _Source._string_pattern is None:
            original = codeanalyze.get_string_pattern()
            pattern = r'(%s)((\s|\\\n|#[^\n]*\n)*(%s))*' % \
                      (original, original)
            _Source._string_pattern = re.compile(pattern)
        repattern = _Source._string_pattern
        return self._consume_pattern(repattern, end)

Here "original" matches a single string literal and
"pattern" matches adjacent string literals that should
be merged; these patterns do handle common cases, but
there may be cases, such as the one you are trying to
fix, which are not covered.  I wonder if "pattern" can
be modified to match the remaining cases.

> This is list comprehension in function args 
> ```python
> fun_call(float(y) for y in x for x in l)
> ````
> 
> As I said, function call can contain almost anything in random order.
> ```python
> call(1, a=1, *[1, "" ""], c = 1, wryyy=lambda r=lambda: x : {'b':1}, **{'b':1})
> ````
> **This can only be parsed by fully functional python language parser. I believe everyone agree with that on this stage.**
> 
> And we have few options:
> * Take cpython's parser, written in C, add needed info, and create cpython fork/library.
> * Take pypy's parser, written in Python. I take pypy and I already know how to patch it.

I just fear that it may take a lot of effort to import that
into rope, especially the difficulty of fixing the files that
traverse ASTs.  However, I admittedly shall be happy to see
that done; it would be no longer necessary for the version of
python running rope itself and the version of python the project
assumes to match.  If the interfaces are identical, we may be
able to have both of them inside rope and switch to the old
behaviour if needed (for performance reasons, for example).

Thanks,
Ali

enomado:
You was right. I take pypy and it was hard to clean it from jit and space related things. Than I take cpython, patch it and it almost works, except hundreds of incorrect nodes.

Here is the result of my sprint:
https://github.com/enomado/cpython
https://github.com/enomado/python-code-region-highlight

As I said, rope is the only tool that can nicely set location of python code for now. Thank you.
Regarding to this pull request - I'll try to clean it later.

mcepl:
What is the status of this PR?

soupytwist:
This seems to be abandoned for over a year now and I am inclined to close it. After reading through I'm still not sure what problem this is trying to solved-- leading with unit tests would make this much clearer. There are still some incompatibilities in `patchedast` though, I will open an issue to track this and hopefully we can come up with some failing test cases.

Thank you. If you do some updates to this, please do reopen and we will review it.

PR #118, opened 2015-07-17
add support for DictComp and SetComp to patched_ast

dwiel:
This fixes a specific bug I had where I could not use extract_method feature if the code being extracted had a set or dict comprehension.  This was due to patched_ast not adding region information to anything in those parts of the ast and so the code could not be reconstructed.

### DIFF
--- a/rope/refactor/patchedast.py
+++ b/rope/refactor/patchedast.py
@@ -493,6 +493,19 @@ class _PatchingASTWalker(object):
         children.append(']')
         self._handle(node, children)
 
+    def _SetComp(self, node):
+        children = ['{', node.elt]
+        children.extend(node.generators)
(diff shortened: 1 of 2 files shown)


dwiel:
Let me know if there is anything else I need to do to help this in

mcepl:
See https://travis-ci.org/python-rope/rope/jobs/71366518 ... tests at least on python 2.* should not fail.

dwiel:
doh, you can guess which version of python I used to test it with :)

dwiel:
ok, there we go.  tests were failing on 2.6 because 2.6 doesn't have set or dict comprehensions.  I check that the python version has them before testing them now.

mcepl:
Thank you.

PR #127, opened 2015-10-01
Remove note about project moving to python-rope/rope

phalgun:
'Cos this project is already at github.com/python-rope/rope


mcepl:
Right. Thank you for reminding me.

PR #130, opened 2015-11-19
Bugfix for parsing strings with extra quote characters

soupytwist:
Hello,
I had an issue with rope crashing or giving truly bizarre results when moving/renaming specific modules, which I tracked down to how rope parses the source files. Any file which contains a multiline string that starts with more than 3 quote characters would cause it to incorrectly compute the logical line numbers for each scope. I've adjusted the RE which is used to delimit scopes and added some new tests for this case.

For example, `""""Hello, world!" example."""` is a valid python string that rope does not handle correctly.

### DIFF
--- a/rope/base/codeanalyze.py
+++ b/rope/base/codeanalyze.py
@@ -129,31 +129,27 @@ class _CustomGenerator(object):
                 i += 1
         return result
 
-    _main_chars = re.compile(r'[\'|"|#|\\|\[|\]|\{|\}|\(|\)]')
+    # Doesn't match quotes which are escaped
+    _main_tokens = re.compile(r'((?<!\\)(\'\'\'|"""|\'|")|#|\[|\]|\{|\}|\(|\))')
 
     def _analyze_line(self, line):
-        char = None
-        for match in self._main_chars.finditer(line):
-            char = match.group()
-            i = match.start()
-            if char in '\'"':
+        token = None
+        for match in self._main_tokens.finditer(line):
+            token = match.group()
+            if token in ["'''", '"""', "'", '"']:
                 if not self.in_string:
-                    self.in_string = char
-                    if char * 3 == line[i:i + 3]:
-                        self.in_string = char * 3
-                elif self.in_string == line[i:i + len(self.in_string)] and \
-                    not (i > 0 and line[i - 1] == '\\' and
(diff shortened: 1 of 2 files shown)


mcepl:
@aligrudi any comments?

aligrudi:
> @aligrudi any comments?

The patch looks good.  Is the effect of the more complex
regular expression on the performance noticeable (I guess
not but just to make sure), for instance for running the
test suite or performing a refactoring on rope itself?

soupytwist:
There is definitely a performance hit on the `_CustomGenerator`-- testing this module individually,parsing a large python file 10,000 times, `custom_generator` takes almost 50% longer. In a more realistic scenario, computing changes to move `codeanalyze.py` to `rope/refactor/`, there was only a slight slowdown (~2%).

I played around with this a bit to see if removing the lookbehind group made a big difference, but it doesn't seem to. If you've got suggestions how to improve, I'd be happy to hear!

aligrudi:
Nick Smith <notifications@github.com> wrote:
> There is definitely a performance hit on the `_CustomGenerator`-- testing this module individually,parsing a large python file 10,000 times, `custom_generator` takes almost 50% longer. In a more realistic scenario, computing changes to move `codeanalyze.py` to `rope/refactor/`, there was only a slight slowdown (~2%).

Thanks for testing.  How about the overall time of parsing
in your test?  Is the slowdown of custom_generator noticeable
compared to the time required for parsing 10000 times?

soupytwist:
Here's some more detail what I tested

**test.py**:
```python
proj = rope.base.project.Project(...)
mod = proj.get_file('rope/base/codeanalyze.py')
dest = proj.get_folder('rope/refactor')
related = [...] # Resources referencing "codeanalyze", 24 total

mv = rope.refactor.move.MoveModule(proj, mod)
changes = mv.get_changes(dest, resources=related)
```

```
time for i in {1..100}; do python test.py; done

Before
real    1m35.227s
user    1m28.000s
sys     0m6.999s

After
real    1m37.924s
user    1m30.564s
sys     0m7.120s
```

Not a big impact, ~25ms longer per run. So, no I don't think the parsing difference is noticeable.

mcepl:
> Not a big impact, ~25ms longer per run. So, no I don't think the parsing difference is noticeable.

I agree.

PR #135, opened 2016-01-05
Don't change kwarg names when moving modules.

voidlily:
This fixes the following error:

```python
rope.base.exceptions.ModuleSyntaxError: Syntax Error in file <...> line <...>: keyword can't be an expression
```

This error comes from having a kwarg in a function with the same name as
the module you're moving. This change fixes this by adding a filter to
ignore keyword parameter names, which is used on the move module
refactor.

### DIFF
--- a/rope/refactor/move.py
+++ b/rope/refactor/move.py
@@ -605,7 +605,8 @@ class _MoveTools(object):
 
     def _create_finder(self, imports):
         return occurrences.create_finder(self.project, self.old_name,
-                                         self.old_pyname, imports=imports)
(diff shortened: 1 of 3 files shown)


mcepl:
+1 from me. @aligrudi , what do you think?

aligrudi:
> +1 from me. @aligrudi , what do you think?

Seems OK.

	Ali

mcepl:
Merged in f62bd24bd3243943fea7a94e31c35468f204e693. Thanks.

PR #149, opened 2016-02-18
Switch to setuptools (`develop` command is kewl).

mcepl:
OK, if I want to make some development on rope I really want ``setuptools`` here (and ``develop`` command). Unless @aligrudi has strong objections, I will merge this into ``devel`` branch.

The only real objection I have heard about it is in http://thread.gmane.org/gmane.comp.python.rope/552/focus=556 which makes a vague statement about "Windows ready made installers" (I am not sure what he was talking about; ``rope`` is used by Python developers who are supposed to be able to use ``pip`` or ``python setup.py``).

@emacsway , @sergeyglazyrindev, any comments?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,12 +1,16 @@
-from distutils.core import setup, Command
 import sys
+
+try:
+    from setuptools import Command, setup
+except ImportError:
+    from distutils.core import Command, setup
 try:
     import unittest2 as unittest
 except ImportError:
     import unittest
 
-
 import rope
+
 import ropetest
 import ropetest.contrib
 import ropetest.refactor


sergeyglazyrindev:
Hey @mcepl, thanks for asking. On one side distutils is pretty outdated and we definitely need to use setuptools or new wheels system but imho we need to remember how rope users are using the package. If eric ide needs distutils, we can leave it and would be great to support some kind of deprecation policy for rope (like django does, let's say for features being deprecated in future releases we can raise DeprecationWarning), let's say if we want to keep it in stable state, we need to think about stability of interface and of course about the stability of the way how to install it
Here's some discussion
http://stackoverflow.com/questions/25337706/setuptools-vs-distutils-why-is-distutils-still-a-thing
http://stackoverflow.com/questions/7537799/setuptools-distutils-installing-files-into-the-distributions-dlls-directory?rq=1
Would be that hard to add a fallback to standart distutils stuff ?
```
try:
    from setuptools import setup
except ImportError:
    from distutils.core import setup
```
Personally for me :) I don't use windows, so it doesn't matter for me :) but on other hand, there's a lot of users who use windows :)

mcepl:
Right you, the top of ``setup.py`` looks awful, but that's just the price to pay for the wide support (2.6-3.5).

sergeyglazyrindev:
Yes, but I see nothing wrong here :)

mcepl:
OK, I see no complaints here, so merging.

PR #152, opened 2016-03-29
Remove all relevant unused imports when moving globals

soupytwist:
Previously only fully-qualified imports of the name are removed, but
this potentially leaves leaves module imports that were only imported
for the name. This patch removes unused imports of the module that
previously contained the name as well.

Ex, moving `mod1.do_something` to another module

Related modules that depend on this function:
```python
from mod1 import do_something  # current implementation will remove
import mod1                    # current implementation will keep
from pkg import mod1           # current implementation will keep
```

I think that `import mod1` or `from pkg import mod1` should also be
removed if they are unused after the global is moved.

Note that this does not affect imports of other modules; only imports
of the name which is being moved or the module containing it are
candidates for deletion.

### DIFF
--- a/rope/refactor/move.py
+++ b/rope/refactor/move.py
@@ -211,6 +211,19 @@ class MoveGlobal(object):
         self.import_tools = self.tools.import_tools
         self._check_exceptional_conditions()
 
+    def _import_filter(self, stmt):
+      module_name = libutils.modname(self.source)
+      if isinstance(stmt.import_info, importutils.NormalImport):
+          return any(module_name == name
+                     for name, alias in stmt.import_info.names_and_aliases)
+      elif isinstance(stmt.import_info, importutils.FromImport):
+          if '.' in module_name:
+              package_name = '.'.join(module_name.split('.')[:-1])
+              if stmt.import_info.module_name == package_name:
+                  return True
+          return stmt.import_info.module_name == module_name
+      return False
+
     def _check_exceptional_conditions(self):
         if self.old_pyname is None or \
            not isinstance(self.old_pyname.get_object(),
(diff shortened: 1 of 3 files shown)


sergeyglazyrindev:
Hey @soupytwist .
That makes sense to me.
Thankks for your contribution
I am not official maintainer of the rope unfortunately
@emacsway , @mcepl 
What do you think ?
It looks good to me

mcepl:
Could you confirm, please, that I have correctly rebased this pull request on top of the current master?

soupytwist:
Hi @mcepl,
Thanks for reviewing it. I'm not sure what you did, but I rebased to master, so should be no conflicts.

Unfortunately I also noticed that I made a mistake in `_source_module_changes` in including the `import_filter` argument in the call to `organize_imports`. The `_import_filter` method is designed to match only imports of the name being moved or the module containing it, but for the source module, this is not imported. Instead, any imports which are unused after moving the block should be removed.

So e.g,
`mod1.py`
```python
import mod2

def func():
  return mod2.do_something()
```

if `func` is moved, the import of `mod2` should also be removed. This is generalized in my change to be _any_ unused imports for simplicity, if you're not okay with this let me know and we can work out a different way. Currently it seems to leave imports alone in the source module, which can leave many unused.

Does it make sense? Sorry for the late update.

mcepl:
> Unfortunately I also noticed that I made a mistake in _source_module_changes in including the import_filter argument in the call to organize_imports. The _import_filter method is designed to match only imports of the name being moved or the module containing it, but for the source module, this is not imported. Instead, any imports which are unused after moving the block should be removed.

So, is this mistake fixed in 2974244 ?

soupytwist:
> So, is this mistake fixed in 2974244 ?

Yes, sorry if this wasn't clear. The change is to remove `import_filter` argument from `organize_imports` call in `_source_module_changes`, plus a couple test updates to match.

mcepl:
Thank you

PR #153, opened 2016-03-29
Add an option to prefer "from pkg import mod" imports

soupytwist:
Rope seemed to always choose the normal import by default, but this
leads to ugly code when the module is deeply nested in pacakges. This
option gives preference to the "from pkg import mod" style from-import.

Ex, when moving `mod1.do_something` to `pkg.mod2`

Before:
`mod3.py`:
```python
import mod1

mod1.do_something()
```

After [`prefer_module_from_imports = False`]:
`mod3.py`
```python
import mod1
import pkg.mod2

pkg.mod2.do_something()
```

After [`prefer_module_from_imports = True`]:
`mod3.py`
```python
import mod1
from pkg import mod2

mod2.do_something()
```

### DIFF
--- a/rope/base/default_config.py
+++ b/rope/base/default_config.py
@@ -79,6 +79,10 @@ def set_prefs(prefs):
     # appear in the importing namespace.
     prefs['ignore_bad_imports'] = False
 
+    # If `True`, rope will insert new module imports as
+    # `from <package> import <module>` by default.
+    prefs['prefer_module_from_imports'] = False
+
     # If `True`, rope will transform a comma list of imports into
     # multiple separate import statements when organizing
     # imports.
--- a/rope/refactor/importutils/__init__.py
+++ b/rope/refactor/importutils/__init__.py
@@ -278,6 +278,7 @@ def add_import(project, pymodule, module_name, name=None):
     imports = get_module_imports(project, pymodule)
     candidates = []
     names = []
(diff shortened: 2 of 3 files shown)


sergeyglazyrindev:
Hey @soupytwist 
It looks good to me, the one thing I worry about is : it took a while to figure out why two almost identical test cases has different result:
https://github.com/python-rope/rope/pull/153/commits/ec291eb32a55be6a6b471529ff769aac2ff96416#diff-8ac4db57b30ae4231435919fa13eed51R66
But then I opened file, found that mod4 is a submodule of the package.
Maybe adding simple comment would help in future developers understand this test ?
I can do it myself for sure once Matej merges your changes.

sergeyglazyrindev:
@emacsway @mcepl 
What do you think guys ?

soupytwist:
Thanks @sergeyglazyrindev, good suggestion. I added a comment to indicate as you suggested and another test which I realized was not covered, the case for inserting a "from pkg import mod" import when prefer_module_from_imports = False, to ensure it doesn't break existing functionality.

mcepl:
I like it. @aligrudi ?

aligrudi:
> I like it. @aligrudi ?

Looks good.

	Ali

mcepl:
Thank you

PR #154, opened 2016-04-07
Increase scrutiny on import filter to avoid removing unrelated imports

soupytwist:
This is an improvement to further specify that any import which would be
examined is one of the following forms (when moving pkg.module.SomeClass):

```python
from pkg import module
from pkg.module import any_name
import pkg.module
```

previously this would have also possibly matched

```python
from pkg import unrelated_module
```

even though this module is not affected.


mcepl:
Merged in a8bbe3e

PR #158, opened 2016-04-29

soupytwist:
Don't make assumptions about the pyname until exceptional conditions are
checked. Previously, if the name was just a variable (not a class or function)
initializing the class would raise an AttributeError (not RefactoringError).

```
  File "/third_party/py/rope/refactor/move.py", line 206, in __init__
    self.old_name = self.old_pyname.get_object().get_name()
AttributeError: 'PyObject' object has no attribute 'get_name'
```

### DIFF
--- a/rope/refactor/move.py
+++ b/rope/refactor/move.py
@@ -203,13 +203,13 @@ class MoveGlobal(object):
         self.project = project
         this_pymodule = self.project.get_pymodule(resource)
         self.old_pyname = evaluate.eval_location(this_pymodule, offset)
+        self._check_exceptional_conditions()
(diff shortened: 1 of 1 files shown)


mcepl:
Actually this shouldn't be merged, my fault. @soupytwist , could I ask for some tests on this, please?

mcepl:
I have reset-ed this commit away. Please, file a new pull request with tests. Thank you and I am sorry for not making it more clear before, that we require tests for bug fixes.

soupytwist:
You are right, thank you. I opened #160, which includes a test for this case for the failure scenario.

Sorry for the delay and missing this the first time. :)

PR #159, opened 2016-05-02

mcepl:
Commit from @soupytwist, unfortunately missing tests and some explanation on reproduction.

Replaces #158


mcepl:
Duplicate of #160

PR #160, opened 2016-05-02

soupytwist:
Don't make assumptions about the pyname until exceptional conditions are
checked. Previously, if the name was just a variable (not a class or function)
initializing the class would raise an AttributeError (not RefactoringError).

```
  File "/third_party/py/rope/refactor/move.py", line 206, in __init__
    self.old_name = self.old_pyname.get_object().get_name()
AttributeError: 'PyObject' object has no attribute 'get_name'
```

\+ simple test to show that this case raises the intended error.


mcepl:
Thank you.

PR #161, opened 2016-05-13
Disable flake8 in default config

koterpillar:
As config.py gets created in the project root, flake8 will, if enabled,
will find it and complain about style violations.

Ignoring .ropeproject in setup.cfg would introduce Rope-specific noise
in every affected project.


mcepl:
Certainly. Thank you.

PR #162, opened 2016-05-31
Support moving global variables with move_global

soupytwist:
This is a refactoring I really find useful, most often in the case of constants which are defined in some file that need to be moved elsewhere. Rope already supports this type of refactoring for functions and classes, but variables are a bit more tricky due to the lack of "def" or other declaration.

This implementation takes a simple approach to move the definition location of a global variable from one module to another, extending the MoveGlobal refactoring. See the added tests for examples.

Please let me know what you think! This case isn't as simple as the others, but I think there's a lot of value in supporting it at least to some degree.


mcepl:
@aligrudi , @emacsway what do you think?

soupytwist:
Ping. Please take a look when you get a chance!

PR #166, opened 2016-07-09
Fix inconsistent use of tabs and spaces in indentation

ranjanashish:


ranjanashish:
I just checked that this is already fixed in `master` but not in `devel`

soupytwist:
Thank you for calling it out! To keep these branches somewhat consistent I've cherry-picked @mcepl's fix from master to this branch instead (there are some other formatting changes here as well).

Closing this, thank you again.

PR #169, opened 2016-07-26
Accounted for starargs and kwargs not being available in python 3.5

DonJayamanne:
- starargs and kwargs are not supported in python 3.5
Modified to maintain backwards compatiblity with safe check using getattr

- fixed indentation of code in unitest


DonJayamanne:
Will submit new PR

PR #170, opened 2016-07-26
Support Starred and KeyWord in Python 3.5 with backwards comptibility

DonJayamanne:
Add support for Starred and keywords (http://greentreesnakes.readthedocs.io/en/latest/nodes.html#Call) due to changes in Python 3.5

Fixes the unit test:
```python
    def test_call_func_and_both_varargs_and_kwargs(self):
        source = 'f(*args, **kwds)\n'
```

This requires the previous PR #169, to ensure unit tests run under Python 3.5

### DIFF
--- a/rope/refactor/patchedast.py
+++ b/rope/refactor/patchedast.py
@@ -481,7 +481,12 @@ class _PatchingASTWalker(object):
         self._handle(node, children)
 
     def _keyword(self, node):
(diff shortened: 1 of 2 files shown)


DonJayamanne:
Build fails as the previous PR #169 is required. 
I haven't mixed the changes, not sure which way you'd want me to go ahead with the PRs.

PR #182, opened 2016-09-29
Use inheritance instead of proxying default filesystem actions

soupytwist:
Enables more code sharing across fscommands implementations.

### DIFF
--- a/rope/base/fscommands.py
+++ b/rope/base/fscommands.py
@@ -58,19 +58,19 @@ class FileSystemCommands(object):
             file_.close()
 
 
-class SubversionCommands(object):
+class SubversionCommands(FileSystemCommands):
 
     def __init__(self, *args):
-        self.normal_actions = FileSystemCommands()
+        super(SubversionCommands, self).__init__()
         import pysvn
         self.client = pysvn.Client()
 
     def create_file(self, path):
-        self.normal_actions.create_file(path)
+        super(SubversionCommands, self).create_file(path)
         self.client.add(path, force=True)
 
     def create_folder(self, path):
-        self.normal_actions.create_folder(path)
+        super(SubversionCommands, self).create_folder(path)
         self.client.add(path, force=True)
 
     def move(self, path, new_location):
@@ -79,15 +79,12 @@ class SubversionCommands(object):
     def remove(self, path):
         self.client.remove(path, force=True)
 
-    def write(self, path, data):
-        self.normal_actions.write(path, data)
 
-
-class MercurialCommands(object):
+class MercurialCommands(FileSystemCommands):
 
     def __init__(self, root):
+        super(MercurialCommands, self).__init__()
         self.hg = self._import_mercurial()
-        self.normal_actions = FileSystemCommands()
         try:
             self.ui = self.hg.ui.ui(
                 verbose=False, debug=False, quiet=True,
@@ -110,11 +107,8 @@ class MercurialCommands(object):
         return mercurial
 
     def create_file(self, path):
-        self.normal_actions.create_file(path)
+        super(MercurialCommands, self).create_file(path)
         self.hg.commands.add(self.ui, self.repo, path)
-
-    def create_folder(self, path):
-        self.normal_actions.create_folder(path)
 
     def move(self, path, new_location):
         self.hg.commands.rename(self.ui, self.repo, path,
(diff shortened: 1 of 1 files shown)


soupytwist:
@mcepl, @aligrudi, can someone please review?

mcepl:
Not sure how it relates, but a run on my RHEL-7 (python 2.7) fails:
```
======================================================================
FAIL: test_time_in_std_mods (ropetest.builtinstest.BuiltinModulesTest)
----------------------------------------------------------------------
  File "/home/matej/archiv/knihovna/repos/rope/ropetest/builtinstest.py", line 503, in test_time_in_std_mods
    self.assertTrue('time' in rope.base.stdmods.standard_modules())
AssertionError: False is not true

======================================================================
FAIL: test_sorting_imports_and_standard_modules2 (ropetest.refactor.importutilstest.ImportUtilsTest)
----------------------------------------------------------------------
  File "/home/matej/archiv/knihovna/repos/rope/ropetest/refactor/importutilstest.py", line 775, in test_sorting_imports_and_standard_modules2
    self.import_tools.sort_imports(pymod))
AssertionError: 'import sys\nimport time\n' != u'import sys\n\nimport time\n'

----------------------------------------------------------------------
Ran 1606 tests in 17.293s

FAILED (failures=2, skipped=4)
```

soupytwist:
Interesting. I don't think it is related to this change, let's open a separate issue for it.

Also, I see your output `Ran 1606 tests` where on my machine runs only 1600 tests-- and the tests which failed for you pass for me, and the same on travis. Are you sure you are working on the right branch? If so maybe there is some problem with pycompat detection.

mcepl:
> Also, I see your output Ran 1606 tests where on my machine runs only 1600 tests-- and the tests which failed for you pass for me, and the same on travis. Are you sure you are working on the right branch? If so maybe there is some problem with pycompat detection.

Here is the output from ``python setup.py clean build test`` in [tests-log.txt](https://github.com/python-rope/rope/files/503716/tests-log.txt).

Also, it is on 5c76244 (commit “Use inheritance instead of proxying default filesystem actions”) on top of 66a5612 (which is my current ``master``).

soupytwist:
Okay. This branch was based a few commits behind master, explains the few extra tests which didn't run. This doesn't affect the tests which failed for you though, so I'm not going to block this PR on it. I opened #183 to investigate your issue... maybe something RHEL-specific?

If no problems with this commit, I'll merge. Thanks for the info!

mcepl:
> explains the few extra tests which didn't run

You are REMOVING tests? Why?

soupytwist:
Nothing is removed! The branch I pushed was based a few commits behind master, so was missing a commit where tests were added. Rebased. Sorry for the misunderstanding!

mcepl:
There is no objection from me (see my comment on #183), but I would like to hear something from @aligrudi .

emacsway:
@soupytwist , as I see, you are done refactoring method "[Replace Delegation with Inheritance](http://refactoring.com/catalog/replaceDelegationWithInheritance.html)". All arguments related to this method are present on page 355 of "Refactoring" book by Martin Fowler (as I see on right site of page by link).

Btw, great article Inheritance vs Composite "[Why extends is evil](http://www.javaworld.com/article/2073649/core-java/why-extends-is-evil.html)".

I'm not sure, that relation between classes SubversionCommands and SubversionCommands has type "[Is a](https://en.wikipedia.org/wiki/Is-a)".

mcepl:
I am not sure how the previous comment relates to this issue (which I can still reproduce with ``master``).

emacsway:
@mcepl , what issue do you mean? Can you give me ticket ID or any other info? My comment related to replace Delegation with Inheritance.

emacsway:
I just give a few arguments advocated delegation, because relation "[Has a](https://en.wikipedia.org/wiki/Has-a)" looks like more appropriate in this case. I think, composition is more appropriate here than inheritance.

mcepl:
I am sorry, I misread this ticket ... it is actually not about https://github.com/python-rope/rope/pull/182#issuecomment-250682940 (or #183). So, it is irrelevant for this ticket.

emacsway:
@soupytwist , suppose, you need a several behaviour for filesystem operations in the future. For example, auto-creation backup files, or case-insensitive mode, or debug mode with logging, or support remote files by sftp/scp etc. How can you do this with inheritance?

soupytwist:
I'm not sure how much of a concern it is for this project, but this change isn't all that important. I'll pass on this.

emacsway:
Sure, it's not a problem in this case. I meant theoretical aspect.

PR #184, opened 2016-10-18
Don't split imports which are not matched by the filter

soupytwist:
This previously would leave the old import in place because it is
readonly; really it shouldn't even touch these imports if they are
supposed to be filtered.


soupytwist:
@mcepl would you take a look please?

mcepl:
I like it (tests pass).

Issue #150, opened 2016-02-19
Make pluggable providers for type hinting


emacsway:

emacsway:
@soupytwist , assign to me, please

emacsway:
Done!
@soupytwist , please, merge #187

mcepl:
OK, is there some piece of documentation what all this means? Could it be somehow used by the plugin writers (e.g., ``ropemacs`` and ``ropevim``)?

emacsway:
@mcepl , it's a good idea. First of all, the pluggable providers required by #146. I need it to go forward. I've added ticket #191 to add docs how to write pluggable type-hinting providers.

Issue #146, opened 2016-02-18
Add type hinting based on annotations PEP-0484


emacsway:

mcepl:
Just to throw here https://github.com/davidhalter/jedi/issues/630#issuecomment-225763004 ... merge requests based on this idea would be probably considered.

@aligrudi , what do you think?

emacsway:
@soupytwist , can you assign this ticket to me, please?

emacsway:
New in Python 3.5:
- https://docs.python.org/3/library/typing.html
- https://hg.python.org/cpython/file/3.5/Lib/typing.py

emacsway:
Blocked by #150 and  #186

emacsway:
#150 is done in #187

PR #187, opened 2016-10-29
#150 Refactoring. Implemented pluggable providers for type hinting.

emacsway:


mcepl:
By swift browsing through the patch, I have only one question: I see just one test. Could we get some more tests to cover for this whole world? Or is the test coverage carried over from the previous commits?

mcepl:
@aligrudi  ??? What do you think?

emacsway:
@mcepl , we test behavior, not implementation. Behavior currently is unchanged. Under refactoring (not re-engineering) we don't have to change the tests.

emacsway:
In future, it will be added providers for PEP-0484 annotations of python3. Also It will be possible to add custom providers by users (using config or [DIP](https://en.wikipedia.org/wiki/Dependency_injection)), for example, for Django models, etc.

Each new behavior will be tested.

emacsway:
> Or is the test coverage carried over from the previous commits?

@mcepl , yes, all behavior covered by [old tests ](https://github.com/emacsway/rope/blob/2cc0fe92688b2d1d7ac9fe7c52fb5305361a4318/ropetest/type_hinting_test.py), I just renamed the module, because for now type-hinting can use not only docstrings.

New behavior will be added under #146

soupytwist:
As @emacsway says, this behavior is already covered by tests, only the implementation has been moved around. Seeing as nothing new is introduced except making it configurable, I don't see any risk in accepting this. Thanks!

emacsway:
We can inline IAttrProvider into IAssignmentProvider.
See https://github.com/emacsway/rope/blob/51ac587ad70e913946190b359813534a4b4a4664/rope/base/oi/soi.py#L79
and https://github.com/emacsway/rope/blob/51ac587ad70e913946190b359813534a4b4a4664/rope/base/oi/soi.py#L84

Let me fix it.

Issue #189, opened 2016-11-22
Inline IAttrProvider into IAssignmentProvider


emacsway:
In continues of #187

We can inline IAttrProvider into IAssignmentProvider.
See https://github.com/emacsway/rope/blob/51ac587ad70e913946190b359813534a4b4a4664/rope/base/oi/soi.py#L79
and https://github.com/emacsway/rope/blob/51ac587ad70e913946190b359813534a4b4a4664/rope/base/oi/soi.py#L84

Let me fix it.

emacsway:
Move this logic into IAssignmentProvider https://github.com/emacsway/rope/blob/51ac587ad70e913946190b359813534a4b4a4664/rope/base/oi/soi.py#L91

emacsway:
And remove IAttrProvider.

emacsway:
Done in #190

PR #190, opened 2016-11-23
#189 Inline IAttrProvider into IAssignmentProvider

emacsway:
Now we have only 3 types of provider.


soupytwist:
LGTM, merging. I know it's difficult for refactoring, but for future, smaller pulls let me review with much more confidence.

Thanks!

emacsway:
Thanks)

Issue #186, opened 2016-10-29
Support for python3.5 typing


emacsway:
- https://docs.python.org/3/library/typing.html
- https://hg.python.org/cpython/file/3.5/Lib/typing.py

emacsway:
@soupytwist , assign to me, please

emacsway:
https://pypi.python.org/pypi/backports.typing

emacsway:
@soupytwist , I think, we have to add support of metaclass... I don't see another way...

emacsway:
That's why this problem was resolved in jedi: https://github.com/davidhalter/jedi/blob/be2a97cd36ce1b7d6f587ef122efc3433b1e8ae2/jedi/evaluate/jedi_typing.py

PR #192, opened 2016-11-26
#186 preparation for support of typing module.

emacsway:
Added parser for type-comments and docstrings, added support for the following syntax:

- Foo
- foo.bar.Baz
- list[Foo] or list[foo.bar.Baz] etc.
- set[Foo]
- tuple[Foo]
- dict[Foo, Bar]
- collections.Iterable[Foo]
- collections.Iterator[Foo]

Nested expressions also allowed:

- collections.Iterable[list[Foo]]

### DIFF
--- a/rope/base/oi/type_hinting/evaluate.py
+++ b/rope/base/oi/type_hinting/evaluate.py
@@ -1,0 +1,353 @@
+# Based on super lightweight Simple Top-Down Parser from http://effbot.org/zone/simple-top-down-parsing.htm
+# and https://bitbucket.org/emacsway/sqlbuilder/src/default/sqlbuilder/smartsql/contrib/evaluate.py
+import re
+from rope.base.utils import pycompat
+from rope.base.oi.type_hinting import utils
+from rope.base import utils as base_utils
+
+
+class SymbolBase(object):
+
+    name = None  # node/token type name
+
+    def __init__(self):
+        self.value = None  # used by name and literals
+        self.first = None
+        self.second = None
+        self.third = None  # used by tree nodes
+
+    def nud(self, parser):
+        raise SyntaxError(
+            "Syntax error (%r)." % self.name
+        )
+
+    def led(self, left, parser):
+        raise SyntaxError(
+            "Unknown operator (%r)." % self.name
+        )
+
+    def evaluate(self, pyobject):
+        raise NotImplementedError(self.name, self)
+
+    def __repr__(self):
+        if self.name == '(name)':
+            return "(%s %s)" % (self.name[1:-1], self.value)
+        out = [repr(self.name), self.first, self.second, self.third]
+        out = [str(i) for i in out if i]
+        return '(' + ' '.join(out) + ')'
+
+
+class SymbolTable(object):
+
+    def multi(func):
+        def _inner(self, names, *a, **kw):
+            for name in names.split():
+                func(self, name, *a, **kw)
+        return _inner
+
+    def __init__(self):
+        self.symbol_table = {}
+
+    def get(self, name, default=None):
+        return self.symbol_table.get(name, default)
+
+    def __getitem__(self, name):
+        return self.symbol_table[name]
+
+    def __iter__(self):
+        return iter(self.symbol_table)
+
+    def symbol(self, name, bp=0):
+        try:
+            s = self.symbol_table[name]
+        except KeyError:
+
+            class S(SymbolBase):
+                pass
+
+            s = S
+            s.__name__ = "symbol-" + name  # for debugging
+            s.name = name
+            s.lbp = bp
+            self.symbol_table[name] = s
+        else:
+            s.lbp = max(bp, s.lbp)
+        return s
+
+    @multi
+    def infix(self, name, bp):
+        symbol = self.symbol(name, bp)
+
+        @method(symbol)
+        def led(self, left, parser):
+            self.first = left
+            self.second = parser.expression(bp)
+            return self
+
+    @multi
+    def infix_r(self, name, bp):
+        symbol = self.symbol(name, bp)
+
+        @method(symbol)
+        def led(self, left, parser):
+            self.first = left
+            self.second = parser.expression(bp - 0.1)
+            return self
+
+    def ternary(self, name, name2, bp):
+        symbol = self.symbol(name, bp)
+        symbol2 = self.symbol(name2)
+
+        @method(symbol)
+        def led(self, left, parser):
+            self.first = left
+            self.second = parser.expression(symbol2.lbp)
+            parser.advance(symbol2.name)
+            self.third = parser.expression(symbol2.lbp + 0.1)
+            return self
+
+    @multi
+    def prefix(self, name, bp):
+        symbol = self.symbol(name, bp)
+
+        @method(symbol)
+        def nud(self, parser):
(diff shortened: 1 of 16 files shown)


rope/base/oi/type_hinting/evaluate.py
```diff
@@ -0,0 +1,393 @@
...
+                break
+            parser.advance(',')
+    parser.advance(')')
+    parser.advance('->')
+    self.third = parser.expression(symbol('->').lbp + 0.1)
+    return self
+
+
+@method(symbol('('))
+def evaluate(self, pyobject):
```
soupytwist: This duplicates the behavior of `SymbolBase`

rope/base/oi/type_hinting/evaluate.py
```diff
@@ -0,0 +1,393 @@
...
+    raise NotImplementedError
+
+
+@method(symbol('or'))
+@method(symbol('|'))
+def evaluate(self, pyobject):
```
soupytwist: Also unnecessary.

rope/base/oi/type_hinting/evaluate.py
```diff
@@ -0,0 +1,393 @@
...
+    def __call__(self, program, pyobject):
+        """Evaluates the program string or AST
+
+        :type program: str or rope.base.oi.type_hinting.evaluate.SymbolBase
+        :rtype: rope.base.pyobjects.PyDefinedObject | rope.base.pyobjects.PyObject or None
+        """
+        ast = self.compile(program) if isinstance(program, pycompat.string_types) else program
+        return ast.evaluate(pyobject)
+
+e = Evaluator()
```
soupytwist: Better name for this? `evaluate` is okay.

ropetest/type_hinting_test.py
```diff
@@ -161,6 +166,105 @@ def test_hint_attr_for_post_defined_type(self):
         result = self._assist(code, offset)
         self.assert_completion_in_result('isAlive', 'attribute', result)
 
+    def test_hint_parametrized_list(self):
+        code = 'class Sample(object):\n' \
```
soupytwist: Can you pull the code out so it's not repeated so many times?

E.g

```python
  _CODE_TEMPLATE = """class Sample(Object):
  {hint}
  def a_method(self):
    for i in self.a_attr:
      i.isA
"""

  def test_hint_parametrized_list(self):
    code = self._CODE_TEMPLATE.format(hint=self._make_class_hint('list[threading.Thread]'))
```

rope/base/oi/type_hinting/evaluate.py
```diff
@@ -0,0 +1,393 @@
...
+            "Syntax error (%r)." % self.name
+        )
+
+    def led(self, left, parser):
+        raise SyntaxError(
+            "Unknown operator (%r)." % self.name
+        )
+
+    def evaluate(self, pyobject):
+        raise NotImplementedError((self.name, vars(self)))
```
emacsway: yes, no problem. But state of object is more important in this case. It's not principled.

rope/base/oi/type_hinting/evaluate.py
```diff
@@ -0,0 +1,393 @@
...
+    @multi
+    def postfix(self, name, bp):
+        symbol = self.symbol(name, bp)
+
+        @method(symbol)
+        def led(self, left, parser):
+            self.first = left
+            return self
+
+    multi = staticmethod(multi)
```
emacsway: No, it's just a decorator defined inside the class. In this place it's already not required. This line is only for stupid code checker in my editor.

rope/base/oi/type_hinting/evaluate.py
emacsway: yes

rope/base/oi/type_hinting/evaluate.py
emacsway: This will be implemented in future.

emacsway: I'll add TODO for explanation

rope/base/oi/type_hinting/evaluate.py
emacsway: ok

rope/base/oi/type_hinting/evaluate.py
```diff
@@ -0,0 +1,393 @@
...
+        )
+
+    def evaluate(self, pyobject):
+        raise NotImplementedError((self.name, vars(self)))
+
+    def __repr__(self):
+        if self.name == '(name)':
+            return "(%s %s)" % (self.name[1:-1], self.value)
+        out = [repr(self.name), self.first, self.second, self.third]
+        out = map(str, filter(None, out))
```
emacsway: No matter

rope/base/oi/type_hinting/evaluate.py
```diff
@@ -0,0 +1,393 @@
...
+
+    token = None
+    next = None
+
+    def __init__(self, lexer):
+        self.lexer = lexer
+
+    def parse(self, program):
+        generator = self.lexer.tokenize(program)
+        try:
```
emacsway: I think no, I can't. First of all, it has return value, but constructor hasn't. Program is argument of state, but lexer is dependency. I create parser by factory method, and factory aware only about dependencies.

emacsway: And this method executes action, it parses the string. But constructor does only object initialization.

rope/base/oi/type_hinting/evaluate.py
```diff
@@ -0,0 +1,393 @@
...
+        def _inner(self, names, *a, **kw):
+            for name in names.split():
+                func(self, name, *a, **kw)
+        return _inner
+
+    def __init__(self):
+        self.symbol_table = {}
+
+    def get(self, name, default=None):
+        return self.symbol_table.get(name, default)
```
emacsway: I think, case insensitive names is useless in our case, I'm removing it.

rope/base/oi/type_hinting/evaluate.py
emacsway: Sorry, this one also not implemented yet, I've added TODO

rope/base/oi/type_hinting/evaluate.py
soupytwist: Ok, I overlooked that this is supplied by an injected dependency. You're right, it's fine.

ropetest/type_hinting_test.py
emacsway: I'm not sure with this proposal.
1. It increases coupling of some methods and decreases cohesion of class.
2. It degrades readability, in case test fails it complicates the understanding and navigation in the code.
3. It's contrary for one of the main purpose of tests - demonstrate case of usage. The demonstration should be as easy as possible.
4. I can't apply this pattern for test_hint_parametrized_nested_tuple_list(), test_hint_parametrized_dict_key() and test_hint_parametrized_dict_value(). So, there is only partial overlap, not systematical.

soupytwist: > I can't apply this pattern for test_hint_parametrized_nested_tuple_list(), test_hint_parametrized_dict_key() and test_hint_parametrized_dict_value().

Oops, I didn't notice this one. Please disregard.

PR #193, opened 2016-11-30
Add an option to disable sorting imports entirely

soupytwist:
Rope, by default, rewrites import statements by removing all the
top-level import lines from the file, then sorting them, then
reinserting them at the top. This behavior of pulling everything to the
top of the file can be undesirable. As an alternative, we can rewrite
the imports in place, which won't sort them, but also won't impose this
extra constraint that all top-level imports must be at the top.

### DIFF
--- a/rope/base/default_config.py
+++ b/rope/base/default_config.py
@@ -89,9 +89,13 @@ def set_prefs(prefs):
     # imports.
     prefs['split_imports'] = False
 
-    # If `True`, rope will sort imports alphabetically by module name
-    # instead of alphabetically by import statement, with from imports
-    # after normal imports.
+    # If `True`, rope will sort the top-level import statements and
+    # reinsert them at the top of the module when making changes.
+    prefs['sort_imports_at_top'] = True
+
+    # If `True` (and `sort_imports_at_top` is also `True`), rope will sort
+    # imports alphabetically by module name instead of alphabetically by import
+    # statement, with from imports after normal imports.
     prefs['sort_imports_alphabetically'] = False
 
     # Location of implementation of rope.base.oi.type_hinting.interfaces.ITypeHintingFactory
--- a/rope/base/default_config.py
+++ b/rope/base/default_config.py
@@ -89,13 +89,13 @@ def set_prefs(prefs):
     # imports.
     prefs['split_imports'] = False
 
-    # If `True`, rope will sort the top-level import statements and
+    # If `True`, rope will remove all top-level import statements and
     # reinsert them at the top of the module when making changes.
-    prefs['sort_imports_at_top'] = True
+    prefs['pull_imports_to_top'] = True
 
-    # If `True` (and `sort_imports_at_top` is also `True`), rope will sort
-    # imports alphabetically by module name instead of alphabetically by import
-    # statement, with from imports after normal imports.
+    # If `True`, rope will sort imports alphabetically by module name instead of
+    # alphabetically by import statement, with from imports after normal
+    # imports.
     prefs['sort_imports_alphabetically'] = False
 
     # Location of implementation of rope.base.oi.type_hinting.interfaces.ITypeHintingFactory
--- a/rope/refactor/importutils/module_imports.py
+++ b/rope/refactor/importutils/module_imports.py
@@ -12,6 +12,7 @@ class ModuleImports(object):
         self.pymodule = pymodule
         self.separating_lines = 0
         self.filter = import_filter
+        self.sorted = False
 
     @property
     @utils.saveit
@@ -50,7 +51,8 @@ class ModuleImports(object):
         return result
 
     def get_changed_source(self):
-        if not self.project.prefs.get("sort_imports_at_top"):
+        if (not self.project.prefs.get("pull_imports_to_top") and
+            not self.sorted):
             return ''.join(self._rewrite_imports(self.imports))
 
         # Make sure we forward a removed import's preceding blank
@@ -211,8 +213,6 @@ class ModuleImports(object):
         return self.pymodule.get_resource().parent
 
     def sort_imports(self):
-        if not self.project.prefs.get("sort_imports_at_top"):
-            return
         if self.project.prefs.get("sort_imports_alphabetically"):
             sort_kwargs = dict(key=self._get_import_name)
         else:
@@ -232,6 +232,7 @@ class ModuleImports(object):
         last_index = self._move_imports(third_party, last_index, 1)
         last_index = self._move_imports(in_projects, last_index, 1)
         self.separating_lines = 2
+        self.sorted = True
 
     def _first_import_line(self):
         nodes = self.pymodule.get_ast().body
--- a/rope/refactor/importutils/module_imports.py
+++ b/rope/refactor/importutils/module_imports.py
@@ -50,6 +50,9 @@ class ModuleImports(object):
         return result
 
     def get_changed_source(self):
+        if not self.project.prefs.get("sort_imports_at_top"):
+            return ''.join(self._rewrite_imports(self.imports))
+
         # Make sure we forward a removed import's preceding blank
         # lines count to the following import statement.
         prev_stmt = None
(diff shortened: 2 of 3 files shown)


soupytwist:
This is a feature I need and I think is generally useful. Rope's current behavior of pulling everything to the top and sorting is useful sometimes, but it's not a necessary step in the refactoring and can create problems, e.g if the import order is important.

I'd like to get another pair of eyes on this at least. @emacsway or @mcepl, would you take a look?

emacsway:
Looks Great!

mcepl:
I wonder whether our current behavior is not determined by [PEP-8](https://www.python.org/dev/peps/pep-0008/). Somebody should make a deep dive into it. I really don't think we should override generally accepted standard.

emacsway:
@mcepl , there is at least two case, when we can't to perform import at the end of file:

1. We need add sys.path before or perform a some initialization.
2. We need resolve circular (or cyclic) imports in Python.

But, maybe, command line argument is more acceptable way than config, I'm not sure...

mcepl:
Yes, I know, it is not easy and I am really not sure what is the proper solution.

soupytwist:
I don't think rope's job should be to make corrections to enforce a style guide. By default, I think it's good to base rope's decisions on this guide, but if the end user does not care about this aspect of the style guide (or there are exceptions) then I think it makes sense to be able to forego parts of it, just the same way that we have "split_imports" as optional.

emacsway:
@soupytwist , as I see, there is already [exists option to enable/disable sort](https://github.com/python-rope/rope/blob/554cfcb1a05293ef5b1773e56650e59f8d9e7a7c/rope/refactor/importutils/__init__.py#L175).

soupytwist:
@emacsway, you're right, but this is a bit misleading. This option does sort the imports, but also any time imports are changed they are pulled to the top and sorted by default. This option would disable both of these cases.

emacsway:
@soupytwist , as I see, there is exists option, but it isn't used, at least in ropemace user interface. Command rope-organize-imports does not ask me about "sort" like rope-rename ask me about "inheritance". Interface simple is not finalized.

emacsway:
@soupytwist , I think, solution of this issue should be here: https://github.com/python-rope/ropemode/blob/8b86f9f542685f5b53449215047b83237aaa14b8/ropemode/refactor.py#L323

emacsway:
@soupytwist , there is the similar situation, and it resolved in interactive way: https://github.com/python-rope/ropemode/blob/8b86f9f542685f5b53449215047b83237aaa14b8/ropemode/interface.py#L254

I think, we need ask user about "sort" option when he call "organize_imports". We should implement already existent option "sort" in user interface instead of adding new config parameter which will disable part of the user interface. 

@mcepl , thanks you for forcing me to dive into the code.

emacsway:
See for example how "rename" method asks user about options: https://github.com/python-rope/ropemode/blob/8b86f9f542685f5b53449215047b83237aaa14b8/ropemode/refactor.py#L126

emacsway:
In the same way rope should ask user about "sort" option.

soupytwist:
This `ImportOrganizer` class you pointed to is not used within other refactorings, only from the `OrganizeImports` refactoring in `ropemode`. Refactors like `move.MoveModule` call `ModuleImports.organize_imports` which [implicitly sorts them](https://github.com/python-rope/rope/blob/bbf365b/rope/refactor/importutils/module_imports.py#L71) and pulls it to the top, which is what I am aiming to prevent.

I think this does actually belong as a configuration preference; it is asking rope to not enforce this part of the style guide. You are right, however, that `OrganizeImports` contradicts this-- so it is an overreach to break from `ModuleImports.sort_imports` if this option is disabled-- I'll remove this change.

Perhaps the option should instead be named `enforce_imports_at_top`, default to `True`?

emacsway:
Rope-mode is an abstract layer between rope and editor (IDE). It's user-interface. It's used by ropemacs and ropevim.

As I see, we [shouldn't use directly ModuleImports.sort_imports](https://github.com/python-rope/rope/search?utf8=%E2%9C%93&q=sort_imports) , but only by [organize_imports](https://github.com/python-rope/rope/blob/554cfcb1a05293ef5b1773e56650e59f8d9e7a7c/rope/refactor/importutils/__init__.py#L173).

It has option "sort" which is equal your [self.project.prefs.get("sort_imports_at_top")](https://github.com/python-rope/rope/pull/193/commits/40a0fd6f97a64424b1aefe8d592fbf9108d30ea5#diff-e819f6a4d36e85e024ec8a24f5e2e2f0R214).

> which implicitly sorts them and pulls it to the top, which is what I am aiming to prevent.

- I think, [it's a bug](https://github.com/python-rope/rope/blame/bbf365bafb4cc49bf0579510e1d5a7042a6466f2/rope/refactor/importutils/module_imports.py#L71), and should be fixed. All sorts should be placed inside ModuleImports.sort_imports(). [SRP](https://en.wikipedia.org/wiki/Single_responsibility_principle).

> I think this does actually belong as a configuration preference; it is asking rope to not enforce this part of the style guide.

- I don't think, for example I need sort in one file, and don't need in other file in the same project. It's option, and this option already has [logical control argument](https://github.com/python-rope/rope/blob/554cfcb1a05293ef5b1773e56650e59f8d9e7a7c/rope/refactor/importutils/__init__.py#L175). But user interface is not aware about this option yet, and also should be fixed.

By the way, sorting of imports and moving (grouping) imports through the file - is the different responsibilities. I think, it's better to separate it.

mcepl:
> @mcepl , thank you for coercing me to dive into the code.

Or you are absolutely welcome!

soupytwist:
Sorry, I misrepresented this; [`organize_imports`](https://github.com/python-rope/rope/blame/bbf365bafb4cc49bf0579510e1d5a7042a6466f2/rope/refactor/importutils/module_imports.py#L71) does not sort in the same way as `sort_imports`. This is not a bug, it's necessary.

The behaviour I am trying to make configurable is that `organize_imports` reinserts the imports at the top of the file, which is useful sometimes, but is not necessary. If the user's code does not abide strictly by PEP-8 style guide, this step is unnecessary and creates noise. This really has nothing to do with sorting! Sorry for the confusion around that.

Here's what I will do:
- Remove `sort_imports_at_top` as this is not the right meaning
- Add configuration option `pull_imports_to_top [= True]` which only affects behavior of `ModuleImports.get_changed_source`
- If `pull_imports_to_top` is `False`, `ModuleImports.get_changed_source` will not remove/reinsert imports at the top of the file unless they are explicitly sorted via `organize_imports`.

And this change really has nothing to do with interactive operation/ropemode.

Does that sound good?

emacsway:
> Sorry, I misrepresented this; organize_imports does not sort in the same way as sort_imports. This is not a bug, it's necessary.

- Do you mean get_changed_source instead of organize_imports?

soupytwist:
> Do you mean get_changed_source instead of organize_imports?

Agh! Yes, sorry. `get_changed_source` does NOT sort the imports. My mistake, thank you for correcting.

emacsway:
> Add configuration option pull_imports_to_top [= True] which only affects behavior of ModuleImports.get_changed_source

- @soupytwist , sounds good. The main problem is that [this methods](https://github.com/python-rope/rope/blame/bbf365bafb4cc49bf0579510e1d5a7042a6466f2/rope/refactor/importutils/module_imports.py#L52) does too many things and has many responsibilities.

I agree, we have to decompose it similar you made it in this PR.

But I think, the more appropriate way is to add option "pull_imports_to_top" or simple "pull_up" [here](https://github.com/python-rope/rope/blob/bbf365bafb4cc49bf0579510e1d5a7042a6466f2/rope/refactor/importutils/__init__.py#L173).

In any case we have to fix [this calling](https://github.com/python-rope/ropemode/blob/8b86f9f542685f5b53449215047b83237aaa14b8/ropemode/refactor.py#L324) for at least "sort" option.

soupytwist:
I pushed another commit to rename and reduce the scope of this to just preventing pulling to the top as discussed.

I'm not sure I agree it belongs as an argument to `ModuleImports.get_changed_source`. This could work, but it only helps if we expect that some calls within a project might want this behavior and others might not. It seems more of a project setting to me, either you want this style forced on your code or you don't. The same argument could be made for [`split_imports`](https://github.com/python-rope/rope/blob/a3c949e/rope/refactor/importutils/__init__.py#L180). I'm not sure it makes sense to plumb these options through every `move` refactoring type when it is a preference that can apply globally.

What do you think?

(I do agree though that `get_changed_source` does too much, I'll take a stab at refactoring this when I get a chance.)

emacsway:
@soupytwist , in any case, I do not raise objections. Your PR is important. This issue is also problem for me, when I'm doing refactoring. You've made the most difficult part, I know how it's hard to work with evaluated tree.

The question about user interface is a trifle.

Unfortunately I'm very busy this week. I would like to try other alternative with UI dialog.
In any case we have to fix [this interface](https://github.com/python-rope/ropemode/blob/8b86f9f542685f5b53449215047b83237aaa14b8/ropemode/refactor.py#L324) to add dialog for "sort" option. And, I think, it's not matter, one or two options we'll have added.

I suggest you to fix [this interface](https://github.com/python-rope/ropemode/blob/8b86f9f542685f5b53449215047b83237aaa14b8/ropemode/refactor.py#L324) before for "sort" option, if you have time this week.

PR #217, opened 2017-10-09
Added support for `async def`.

abingham:
This includes a basic unit test, though it might be good to extend all of the
current function rename tests to test async as well.

### DIFF
--- a/rope/base/pyobjectsdef.py
+++ b/rope/base/pyobjectsdef.py
@@ -354,6 +354,9 @@ class _ScopeVisitor(object):
             self.names[node.name] = pynames.DefinedName(pyfunction)
         self.defineds.append(pyfunction)
 
+    def _AsyncFunctionDef(self, node):
+        return self._FunctionDef(node)
+
     def _Assign(self, node):
         ast.walk(node, _AssignVisitor(self))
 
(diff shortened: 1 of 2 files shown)


soupytwist:
Thank you for creating a fix. Can you please add to the test case: `@testutils.only_for('3.5')` (ropetest.testutils).

emacsway:
Oh, I see that you've created the complete fix. Let me to think a little about the fix. Thank you, @abingham

rope/base/pyobjectsdef.py
```diff
@@ -354,6 +354,9 @@ def _eval(type_=type_, arg=arg):
             self.names[node.name] = pynames.DefinedName(pyfunction)
         self.defineds.append(pyfunction)
 
+    def _AsyncFunctionDef(self, node):
+        return self._FunctionDef(node)
```
emacsway: This looks ok.

> An async def function definition. Has the same fields as FunctionDef

http://greentreesnakes.readthedocs.io/en/latest/nodes.html?highlight=AsyncFunctionDef#async-and-await

emacsway:
I have no objections. Looks good. Please, don't forget add @testutils.only_for('3.5') to fix tests for Python2, as it was said by @soupytwist .

And we still need add support for Await, AsyncFor, AsyncWith, and YieldFrom. But it's not related to this ticket.

abingham:
I've added the version check decorator. I'll also see about putting together a second PR with support for the other async bits.

abingham:
@emacsway #218 includes support for the rest of the async nodes.

soupytwist:
Thank you @abingham, sorry for the delay in closing this.

abingham:
No worries. Glad I could pitch in!

PR #218, opened 2017-10-10
Complete async

abingham:
This builds upon [the `async def` PR](https://github.com/python-rope/rope/pull/217) to add support for all of the new async-related nodes.

A few notes:

- It didn't seem necessary to do anything with the `Await` node, so I didn't add it to the AST walker.
- I'm not entirely sure if aliasing `_With` for `_AsyncWith` is correct. The `with` handler is doing *something* with `__enter__`, but the equivalent for `async with` would be `__aenter__`. Everything seems to work, but a more expert eye would certainly be welcome.

### DIFF
--- a/rope/base/pyobjectsdef.py
+++ b/rope/base/pyobjectsdef.py
@@ -354,6 +354,9 @@ class _ScopeVisitor(object):
             self.names[node.name] = pynames.DefinedName(pyfunction)
         self.defineds.append(pyfunction)
 
+    def _AsyncFunctionDef(self, node):
+        return self._FunctionDef(node)
+
     def _Assign(self, node):
         ast.walk(node, _AssignVisitor(self))
 
--- a/rope/base/pyobjectsdef.py
+++ b/rope/base/pyobjectsdef.py
@@ -404,6 +404,9 @@ class _ScopeVisitor(object):
         for child in node.body:
             ast.walk(child, self)
 
+    def _AsyncWith(self, node):
+        return self._With(node)
+
     def _excepthandler(self, node):
         node_name_type = str if pycompat.PY3 else ast.Name
         if node.name is not None and isinstance(node.name, node_name_type):
--- a/rope/base/pyobjectsdef.py
+++ b/rope/base/pyobjectsdef.py
@@ -369,6 +369,9 @@ class _ScopeVisitor(object):
         for child in node.body + node.orelse:
             ast.walk(child, self)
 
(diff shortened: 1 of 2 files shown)


emacsway:
+ @sergeyglazyrindev

emacsway:
> I'm not entirely sure if aliasing _With for _AsyncWith is correct. 

"async for loops and async with context managers. They have the same fields as For and With, respectively." http://greentreesnakes.readthedocs.io/en/latest/nodes.html?highlight=AsyncFunctionDef#AsyncWith

emacsway:
> I'm not entirely sure if aliasing _With for _AsyncWith is correct. rope.base.pyobjectsdef._ScopeVisitor._With is doing something with __enter__, but the equivalent for async with would be __aenter__. Everything seems to work, but a more expert eye would certainly be welcome.

Yes, it looks like at least this line is affected: https://github.com/emacsway/rope/blob/7d265872676cec114f1a709c978d82b4309bec5a/rope/base/pyobjectsdef.py#L397

emacsway:
As I see, there is also __aiter__() and __anext__() methods. These affect this line: https://github.com/emacsway/rope/blob/7d265872676cec114f1a709c978d82b4309bec5a/rope/base/pyobjectsdef.py#L365

emacsway:
> rope.base.pyobjectsdef._ClassInitVisitor

Looks like yes, but I need some time to investigate it. https://stackoverflow.com/questions/33128325/how-to-set-class-attribute-with-await-in-init

emacsway:
> What about _SuiteWalker and _PatchingASTWalker?

I'm not ready to answer right now. Probably yes. I need a little time to investigate the matter.

abingham:
Regarding `__aenter__`, `__aexit__`, and `__aiter__`, should I just make changes that seem analogously correct? For example, should `_AsyncWith()` be just like `_With()` but with `__aenter__` instead of `__enter__`? The truth is that I don't really know what `_update_evaluated` is doing, so I'd just be cargo-culting without some more explanation/guidance. I didn't use `__aenter__` to begin with since`_AsyncWith` seems to work just fine without it.

emacsway:
@abingham , I'll answer a little later, I was too busy today.

abingham:
And chance someone could look at this?

mcepl:
OK, @abingham if you fix the conflicts soon enough, I am going to merge this on next Monday (February 10), unless @emacsway says otherwise.

PR #221, opened 2017-10-21
Update setup.py

cedspam:
improve python 3 support

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,5 @@
 import sys
-
+sys.path.append(".")
 try:
     from setuptools import Command, setup
 except ImportError:


soupytwist:
Hi, can you please provide some context? Why is this change necessary and what problem is it solving?

Thanks

cedspam:
I had carmap decoding errors without this change, some of the character
codes aren't available in all encoding and over 127 codes have different
meaning for different encodings.
In fact not specifying encoding while reading files with non base ascii
characters is not portable and depends on locale settings
I had to guess the correct encoding and utf8 was logical.
It may be a linked also to a problem in the way in which json encodes
strings with special characters

mcepl:
> I had carmap decoding errors without this change,

Could you please explain to me how adding current directory to ``sys.path`` helps in decoding issues? I don't understand.

mcepl:
@cedspam ping?

cedspam:
sorry, for answering late and having answered off subject. Adding local dir to the search path solve some problems concerning import from local search path you may have on python3

mcepl:
“some problems concerning import from local search path” still doesn’t explain to me why should we open sys.path to random directory where you run rope from (we are not talking about the directory where rope is located), which is a security risk.

Closing this as IHAVENOIDEAWHATSGOINGON. If you want to reopen, please, do so, but provide as with sufficient information:

* what was the goal you wanted to achieve
* what you did (exact steps)
* what you did hope to happen
* what actually happened (including all error messages, tracebacks, logs, etc. in full)

Thank you and I am sorry we are not able to help you better.

PR #227, opened 2017-11-23
Fix PEP-8 errors in rope/base/default_config.py

OddBloke:
default_config.py is written in to .ropeproject within projects using
rope.  This can cause linting checks for the project to start failing
due to the PEP-8 errors that are fixed herein.

These fixes were based on errors reported by pycodestyle (née pep8).


soupytwist:
Thanks!

PR #230, opened 2018-01-03
Bugfix for detecting pynames in aliased import statements

### LINKED ISSUE #229: Rename of module does not handle import aliases correctly
I am using rope to rename modules in my project and came across a problem with how it handles aliases. Here is a minimal example that shows this behavior. In this case, `mod2.py` imports `mod1.py` and so should import `mod1.py` with a new name when `mod1.py`'s name is changed. However, in this case no changes are actually made in `mod2.py`. I am using version 0.10.7.

Here is my example:
```python
from rope.refactor.rename import Rename

project = Project('.')

mod1 = project.root.create_file('mod1.py')
mod1.write(
    'class Test(object):\n'
    '    def __init__(self):'
    '        print "hello world"')

mod2 = project.root.create_file('mod2.py')
mod2string = 'import mod1 as m\n' + 'm.Test()'
mod2.write(mod2string)

new_name = 'AAAAAAAA'
changes = Rename(project, mod1).get_changes(new_name)
project.do(changes)

expected = 'import ' + new_name + ' as m\n' + 'm.Test()'
with open('mod2.py', 'r') as f:
    actual = f.read()
print '~' * 50
print actual
print '~' * 50
print expected
assert  actual == expected

```
I get the following output
```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
import mod1 as m
m.Test()
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
import AAAAAAAA as m
m.Test()

---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-52-b92f47d557f0> in <module>()
     25     print '~' * 50
     26     print expected
---> 27     assert  actual == expected
     28 

AssertionError: 
```

soupytwist:
This case wasn't being handled correctly; e.g, the pyname for "foo" in
`import foo as bar` evaluates to an `AssignedName` instead of an
`ImportedModule`. This is similar to the module name in a "from" import,
adding a special case to handle it in the same way.

Fixes #229

### DIFF
--- a/rope/base/evaluate.py
+++ b/rope/base/evaluate.py
@@ -101,8 +101,9 @@ class ScopeNameFinder(object):
                                                      offset, lineno):
             name = self.worder.get_primary_at(offset).strip()
             return (None, holding_scope.parent[name])
-        # from statement module
-        if self.worder.is_from_statement_module(offset):
+        # module in a from statement or an imported name that is aliased
+        if (self.worder.is_from_statement_module(offset) or
+            self.worder.is_import_statement_aliased_module(offset)):
             module = self.worder.get_primary_at(offset)
             module_pyname = self._find_module(module)
             return (None, module_pyname)
--- a/rope/base/worder.py
+++ b/rope/base/worder.py
@@ -65,6 +65,9 @@ class Worder(object):
     def is_from_aliased(self, offset):
         return self.code_finder.is_from_aliased(offset)
 
+    def is_import_statement_aliased_module(self, offset):
+        return self.code_finder.is_import_statement_aliased_module(offset)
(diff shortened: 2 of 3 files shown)


soupytwist:
@mcepl, want to take a look? Good to have another pair of eyes on it.

mcepl:
Looks reasonable. I cannot vouch for the code, because I really forgot most of Rope API, but I don't see anything obviously stupid and there are tests, so it should be OK.

PR #236, opened 2018-03-14
Generating class in other module.

dryobates:
It adds ability to generate class in module other than current.

### DIFF
--- a/rope/contrib/generate.py
+++ b/rope/contrib/generate.py
@@ -42,15 +42,16 @@ def create_package(project, name, sourcefolder=None):
 
 class _Generate(object):
 
-    def __init__(self, project, resource, offset):
+    def __init__(self, project, resource, offset, goal_resource=None):
         self.project = project
         self.resource = resource
+        self.goal_resource = goal_resource
         self.info = self._generate_info(project, resource, offset)
         self.name = self.info.get_name()
         self._check_exceptional_conditions()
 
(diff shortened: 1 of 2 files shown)


rope/contrib/generate.py
```diff
@@ -42,15 +42,16 @@ def create_package(project, name, sourcefolder=None):
 
 class _Generate(object):
 
-    def __init__(self, project, resource, offset):
+    def __init__(self, project, resource, offset, goal_resource=None):
```
soupytwist: It should be invalid for `GenerateModule` or `GeneratePackage`, there's nothing additional to do for these. However, being as they override `get_changes`, maybe this doesn't matter since it would just be ignored.

dryobates: I admit that it is poor style to declare fields in parent class that are not used in child class. I have chosen it because it needs the smallest changes.

As an alternative I could change hierarchy so that it would look something like:
- _ModuleElementGenerate(_Generate)
- GenerateClass(_ModuleElementGenerate)
- GenerateFunction(_ModuleElementGenerate)
- GenerateVariable(_ModuleElementGenerate)
- GenerateModule(_Generate)
- GeneratePackage(_Generate)

Should I change it?

soupytwist: No, I agree it's better as-is, and won't cause any harm. Thanks!

PR #246, opened 2018-06-23
Avoid overlapping regions when extract similar=True

brendanator:
If similar regions are found during extract method that overlap, the refactored code is invalid

For example, if the first two lines of `func` are extracted
```python
def func(p):
    a = p
    b = a
    c = b
    d = c
    return d
```
the following is produced
```python
def func(p):
    b = new_func(p)c = new_func(a)d = new_func(b)
    b = new_func(p)
    d = new_func(b)
    return d
def new_func(p):
    a = p
    b = a
    return b
```

With this PR a correct refactoring is performed
```python
def func(p):
    b = new_func(p)
    d = new_func(b)
    return d
def new_func(p):
    a = p
    b = a
    return b
```

### DIFF
--- a/rope/refactor/extract.py
+++ b/rope/refactor/extract.py
@@ -251,8 +251,15 @@ class _ExtractPerformer(object):
         finder = similarfinder.SimilarFinder(self.info.pymodule)
         matches = []
         for start, end in regions:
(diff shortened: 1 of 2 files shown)


soupytwist:
Looks good, thank you for debugging + fixing this!

PR #251, opened 2018-08-06
Perform signature verification on pickled data transferred over sockets

### LINKED ISSUE #105: CVE-2014-3539 python-rope: pickle.load of remotely supplied data with no authentication
https://bugzilla.redhat.com/show_bug.cgi?id=1116485

soupytwist:
Before unpickling anything, ensure that it has a valid digital
signature using a randomly-generated shared key. In order for an attacker
to send or tamper with data on the same socket, they must know this key
to compute a valid signature.

Fixes #105 by requiring some authentication before unpickling.

### DIFF
--- a/rope/base/oi/doa.py
+++ b/rope/base/oi/doa.py
@@ -15,8 +15,16 @@ import threading
 
 
 def _compat_compare_digest(a, b):
+    """Implementation of hmac.compare_digest for python < 2.7.7.
+
+    This function uses an approach designed to prevent timing analysis by
+    avoiding content-based short circuiting behaviour, making it appropriate
+    for cryptography.
+    """
     if len(a) != len(b):
         return False
+    # Computes the bitwise difference of all characters in the two strings
+    # before returning whether or not they are equal.
     difference = 0
     for (a_char, b_char) in zip(a, b):
         difference |= ord(a_char) ^ ord(b_char)
--- a/rope/base/oi/doa.py
+++ b/rope/base/oi/doa.py
@@ -2,9 +2,9 @@ import base64
 import hashlib
 import hmac
 try:
-    import pickle
-except ImportError:
     import cPickle as pickle
+except ImportError:
+    import pickle
 import marshal
 import os
 import socket
@@ -12,7 +12,6 @@ import subprocess
 import sys
 import tempfile
 import threading
-import uuid
 
 
 class PythonFileRunner(object):
@@ -118,11 +117,11 @@ class _SocketReceiver(_MessageReceiver):
     def __init__(self):
         self.server_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.data_port = 3037
-        self.key = uuid.uuid4().bytes
+        self.key = bytes(os.urandom(32))
 
         while self.data_port < 4000:
             try:
-                self.server_socket.bind(('127.0.0.1', self.data_port))
+                self.server_socket.bind(('localhost', self.data_port))
                 break
             except socket.error:
                 self.data_port += 1
--- a/rope/base/oi/doa.py
+++ b/rope/base/oi/doa.py
@@ -14,6 +14,20 @@ import tempfile
 import threading
 
 
+def _compat_compare_digest(a, b):
+    if len(a) != len(b):
+        return False
+    difference = 0
+    for (a_char, b_char) in zip(a, b):
+        difference |= ord(a_char) ^ ord(b_char)
+    return difference == 0
+
+try:
+    from hmac import compare_digest
+except ImportError:
+    compare_digest = _compat_compare_digest
+
+
 class PythonFileRunner(object):
     """A class for running python project files"""
 
@@ -117,7 +131,7 @@ class _SocketReceiver(_MessageReceiver):
     def __init__(self):
         self.server_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.data_port = 3037
-        self.key = bytes(os.urandom(32))
+        self.key = os.urandom(32)
 
         while self.data_port < 4000:
             try:
@@ -160,7 +174,7 @@ class _SocketReceiver(_MessageReceiver):
                     continue
 
                 digest = hmac.new(self.key, buf_data, hashlib.sha256).digest()
-                if buf_digest != digest:
+                if not compare_digest(buf_digest, digest):
                     # Signature mismatch; the payload cannot be trusted and just
                     # has to be dropped. See CVE-2014-3539.
                     continue
--- a/rope/base/oi/doa.py
+++ b/rope/base/oi/doa.py
@@ -1,3 +1,6 @@
+import base64
+import hashlib
+import hmac
 try:
     import pickle
 except ImportError:
@@ -9,6 +12,7 @@ import subprocess
 import sys
 import tempfile
 import threading
+import uuid
 
 
 class PythonFileRunner(object):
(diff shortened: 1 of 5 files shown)


rope/base/oi/doa.py
```diff
@@ -114,24 +118,55 @@ class _SocketReceiver(_MessageReceiver):
     def __init__(self):
         self.server_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.data_port = 3037
+        self.key = uuid.uuid4().bytes
```
ssbr: could just use `os.urandom(16)` and it'd be more obviously correct. (It's not obvious without reading its source that uuid4 uses a CSPRNG, although it "should", and does.)

rope/base/oi/doa.py
```diff
@@ -114,24 +118,55 @@ class _SocketReceiver(_MessageReceiver):
...
+
         while self.data_port < 4000:
             try:
-                self.server_socket.bind(('', self.data_port))
+                self.server_socket.bind(('127.0.0.1', self.data_port))
```
ssbr: can you use localhost or similar for better ipv6 something?

rope/base/oi/doa.py
```diff
@@ -114,24 +118,55 @@ class _SocketReceiver(_MessageReceiver):
...
+                    buf_digest = base64.b64decode(buf[:digest_end])
+                    buf_data = buf[digest_end + 1:-1]
+                    decoded_buf_data = base64.b64decode(buf_data)
+                except:
+                    # Corrupted data; the payload cannot be trusted and just has
+                    # to be dropped. See CVE-2014-3539.
+                    continue
+
+                digest = hmac.new(self.key, buf_data, hashlib.sha256).digest()
+                if buf_digest != digest:
```
ssbr: use [`hmac.compare_digest`](https://docs.python.org/3/library/hmac.html#hmac.compare_digest) to avoid timing attacks. Otherwise people can discover the correct hmac for their malicious input by continually sending requests and seeing how long it takes to be rejected.

ropetest/doatest.py
```diff
@@ -0,0 +1,85 @@
+import base64
+import hashlib
+import hmac
+import multiprocessing
+try:
+    import pickle
+except ImportError:
```
ssbr: other way around. pickle will always succeed, it's `cPickle` that might not depending on Python version / implementation.

So it should be:

```python
try:
    import cPickle as pickle
except ImportError:
    import pickle
```

rope/base/oi/doa.py
ssbr: In fact, I think usually people prefer to use the digest size as the key size, so 16 is half as big as it should be. I'd replace with `os.urandom(32)`. (Without making any statement about whether 128 bits of security is too few or whatever). Trying to remember where I heard this but can't find it.

ropetest/doatest.py
```diff
@@ -0,0 +1,85 @@
...
+        digest = hmac.new(b'invalid-key', pickled_data, hashlib.sha256).digest()
+        payload = (base64.b64encode(digest) + b':' +
+                   base64.b64encode(pickled_data) + b'\n')
+        received_objs = self.try_CVE_2014_3539_exploit(receiver, payload)
+
+        # Make sure the exploit did not run
+        self.assertEqual(0, len(received_objs))
+        
+    def test_CVE_2014_3539_sanity(self):
+        # Tests that sending valid, encrypted data on the socket does work.
```
ssbr: signed, not encryoted

rope/base/oi/doa.py
soupytwist: Much nicer, thank you.

rope/base/oi/doa.py
soupytwist: localhost might work.

rope/base/oi/doa.py
soupytwist: I see, but this is only available for python 2.7+ (rope supports as low as 2.6). I could use the stronger protection if available, but I don't think there's much possibility of this attack because 1) the attacker doesn't get feedback of whether their attempt is successful and 2) the process is very short-lived so any brute-force style attack like this would take too long.

ropetest/doatest.py
soupytwist: Huh, I copied this from elsewhere without noticing. Yeah, that's not right!

ropetest/doatest.py
soupytwist: Artifact from a previous idea, ty  :)

rope/base/oi/doa.py
ssbr: > the attacker doesn't get feedback of whether their attempt is successful

I think this is untrue. The longer it takes for the connection to be closed, the more successful they have been.

> the process is very short-lived so any brute-force style attack like this would take too long.

This may be so, but it's a precondition that might change later.

In general I don't like security-critical code that is "delicate" like this. If we ever had a long-lived connection, would anyone think to fix this code to restore its security properties? I doubt it.  We can and should use a constant time comparison and avoid complex contextual guarantees / cleverness. This is probably how I'd write it:

```python
def _compat_compare_digest(a, b):
    if len(a) != len(b): return False
    if sys.version_info.major <= 2:
        a = map(ord, a)
        b = map(ord, b)
    accum = 0
    for (a_char, b_char) in zip(a, b):
        accum |= a_char ^ b_char
    return accum == 0

try:
    from hmac import compare_digest
except ImportError:
    compare_digest = _compat_compare_digest
```

Test case:

```python
    def test_eq_empty(self):
        self.assertTrue(doa._compat_compare_digest('', ''))


    def test_eq_nonempty(self):
        self.assertTrue(doa._compat_compare_digest('abc', 'abc'))


    def test_neq_samelength(self):
        self.assertTrue(doa._compat_compare_digest('abc', 'abd'))


    def test_neq_difflength(self):
        self.assertTrue(doa._compat_compare_digest('abc', 'abcd'))
```


It's much easier to drop support for 2.6. Who uses 2.6, anyway?

rope/base/oi/doa.py
```diff
@@ -118,11 +117,11 @@ class _SocketReceiver(_MessageReceiver):
     def __init__(self):
         self.server_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.data_port = 3037
-        self.key = uuid.uuid4().bytes
+        self.key = bytes(os.urandom(32))
```
ssbr: `os.urandom` always returns bytes in all python versions

soupytwist:
```
>>> import os
>>> type(os.urandom(32))
<type 'str'>
```

ah, I forgot there's no distinction in python2. Done.

rope/base/oi/doa.py
soupytwist: I don't know who uses 2.6, or even that it's entirely supported (but the tests do run and pass).

rope/base/oi/doa.py
```diff
@@ -114,24 +131,55 @@ class _SocketReceiver(_MessageReceiver):
...
-                self.server_socket.bind(('', self.data_port))
+                self.server_socket.bind(('localhost', self.data_port))
                 break
             except socket.error:
                 self.data_port += 1
         self.server_socket.listen(1)
 
     def get_send_info(self):
-        return str(self.data_port)
+        return '%d:%s' % (self.data_port,
```
mcepl: If you were using ``random.SystemRandom()``, you could use its methods, for example:
```python
rd = random.SystemRandom()
rand_str = {:X}'.format(rd.getrandbits(25))
```
with slightly more predictable and manageable results.

rope/base/oi/doa.py
```diff
@@ -11,6 +14,20 @@
 import threading
 
 
+def _compat_compare_digest(a, b):
```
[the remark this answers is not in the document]
soupytwist: I don't know that it is crucial, but deprecating an old version to support this function doesn't seem worthwhile imo (it's pretty simple). I will document it, though, good point.

rope/base/oi/doa.py
soupytwist: I don't think you meant this line, but I'm not sure how it's different than `urandom` that I am using now. That seems just as simple/manageable to me for this case.

ssbr: I'm not sure I understand the suggestion here. `random.SystemRandom` is just a wrapper around `urandom` that lets you get other things like random integers, floats, etc. -- but if you just want a random string, `urandom` is easier.

rope/base/oi/doa.py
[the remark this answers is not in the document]
ssbr: FWIW the bit math here is actually pretty standard (I wrote it from memory). It's still not necessarily constant time but what can you do?

PR #255, opened 2018-09-19
Properly close file handle upon reading contents

hpapaxen:


soupytwist:
Would make more sense to use a `with` statement, but it's an improvement. Thanks!

hpapaxen:
Ah you're right. I'll keep that in mind for future PRs. Thanks and glad to help :-)

PR #210, opened 2017-05-17
fix installation issue on windows

tzing:
install from pip would failed on my windows
```
(tmpenv) λ pip install rope
Collecting rope
  Downloading rope-0.10.5.tar.gz (243kB)
    100% |████████████████████████████████| 245kB 2.1MB/s
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "C:\Users\tzu-ting\AppData\Local\Temp\pip-build-j1x3c_at\rope\setup.py", line 67, in <module>
        long_description=get_long_description(),
      File "C:\Users\tzu-ting\AppData\Local\Temp\pip-build-j1x3c_at\rope\setup.py", line 60, in get_long_description
        lines = open('README.rst').read().splitlines(False)
    UnicodeDecodeError: 'cp950' codec can't decode byte 0xc4 in position 407: illegal multibyte sequence

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in C:\Users\tzu-ting\AppData\Local\Temp\pip-build-j1x3c_at\rope\
```

just add `encoding='utf8'` to the line and it works

PR #259, opened 2019-02-06
fix installation issue on windows

mcepl:
This is a replacement for lost #210.


mcepl:
@emacsway , @abingham , @sergeyglazyrindev ... what I said about merging this on Monday still stands. If you have any problem with this PR, please, speak now and forever hold your peace.

PR #263, opened 2019-02-28
Setup version

mcepl:
I have this bug report in [email](https://groups.google.com/d/msg/rope-dev/afB5KgNhSvk/zkubyRUZAwAJ):
```
OS: windows 10
Python: 3.6.7

When I install rope, I have this error: 

Collecting rope
  Using cached 
https://files.pythonhosted.org/packages/af/9b/e92c1d561631da9fbeaf4d5d67ecb65b7d284a63069ee37aec44a2eefae4/rope-0.12.0.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File 
"C:\Users\abukreev\AppData\Local\Temp\pip-install-jve2b1hl\rope\setup.py", 
line 13, in <module>
        import rope
    ModuleNotFoundError: No module named 'rope'
```

I agree that importing too much in setup.py is a bad idea.

What do you think?

### DIFF
--- a/rope/__init__.py
+++ b/rope/__init__.py
@@ -1,7 +1,7 @@
 """rope, a python refactoring library"""
 
 INFO = __doc__
-VERSION = '0.11.0'
+VERSION = '0.12.0'
 COPYRIGHT = """\
 Copyright (C) 2015-2018 Nicholas Smith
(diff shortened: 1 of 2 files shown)


soupytwist:
LGTM, thanks for fixing!

PR #266, opened 2019-03-12
Relicensing to LGPLv3+

### LINKED ISSUE #262: Relicense rope under LGPL
Given https://github.com/Microsoft/vscode-python/issues/3867 I would like to consider relicensing whole ``rope`` project under another license more accessible to editors/IDE desiring to use rope.

It is not only attempt to butter VS Code developers up, but truly plain GPL doesn't make sense much for the project, which is essentially a library. Despite what [some others say](https://www.gnu.org/philosophy/pragmatic.html), I don't think tricking somebody into free software by spreading infection is not The Right Thing to do, and so I would prefer to relicense under an non-infectious license.

The question remains which license to use. In order to lessen friction (and because I like it), I would suggest [LGPL3+](https://www.gnu.org/licenses/lgpl.html) because it is the up-to-date version closest to the current license, and thus hopefully least controversial.

I would like to ask all you listed below add a comment with your agreement (just +1 is enough, but please make a separate comment so we have record of your name) if you agree with this change to LGPL3+. If you don't agree or if you really want a different license, please, make a comment as well. Of course, I give my +1 by filing this issue now.

This list should be hopefully list of all people who have more than two commits in the tree. If you are a committer and you are not listed below, please, let us know as well.

Thank you.

- [ ] @abingham
- [ ] @aligrudi
- [ ] @brendanator
- [ ] @digenis
- [x] @dsyzling
- [x] @emacsway
- [x] @enomado
- [x] @ferada
- [x] @gwelymernans
- [x] @hayd
- [x] @hpapaxen
- [x] @jorgenschaefer
- [ ] @MarekLew
- [x] @mcepl 
- [x] @orestis
- [x] @piranha
- [x] @sergeyglazyrindev
- [x] @soupytwist
- [x] @stpierre
- [x] @tzing

mcepl:
Fixes #262

Please, review and comment swiftly.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,339 +1,165 @@
-		    GNU GENERAL PUBLIC LICENSE
-		       Version 2, June 1991
+                   GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
 
- Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
- 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
-			    Preamble
 
-  The licenses for most software are designed to take away your
-freedom to share and change it.  By contrast, the GNU General Public
-License is intended to guarantee your freedom to share and change free
-software--to make sure the software is free for all its users.  This
-General Public License applies to most of the Free Software
-Foundation's software and to any other program whose authors commit to
-using it.  (Some other Free Software Foundation software is covered by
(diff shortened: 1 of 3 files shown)


soupytwist:
Matej, I'm sorry for the slowness on this. I have been on vacation for the past week and fairly disconnected. I would have preferred to receive explicit permission from @aligrudi before going ahead with this, but seeing as he is unresponsive and you have issued many notices about this, I think we only can proceed with this. Thank you for your effort.

mcepl:
@soupytwist I was slow to react. I was harshly promising merge without mercy and then I got distracted :(. Yes, I would love @aligrudi 's reaction as well, but he apparently got lost in the bowls of history. If he sues us, we will have to rewrite good part of rope, but I hope he won't.

mcepl:
So, we should release now, shouldn't we?

soupytwist:
0.14.0 is released.

PR #267, opened 2019-03-20
Properly map 'open' to builtin function (as opposed to type)

hpapaxen:


hpapaxen:
@soupytwist I also have a fix for recognizing type annotations in python 3. i.e.
```
with open('file') as f:
        a: str = f.read()
```

If you'd like I can issue a PR for that too.

rope/base/builtins.py
```diff
@@ -587,17 +589,21 @@ def get_returned_object(self, args):
 
 class File(BuiltinClass):
 
-    def __init__(self):
+    def __init__(self, filename=None, mode='r', *args):
```
soupytwist: It looks like these extra arguments are never used. Are they needed?

hpapaxen: Yeah they're needed because of [open](https://docs.python.org/2/library/functions.html#open). For example, if I do:
```
with open("file") as f:
    pass
```
I'll get: `TypeError: __init__() takes 1 positional argument but 2 were given`

soupytwist:
> @soupytwist I also have a fix for recognizing type annotations in python 3. i.e.
> 
> ```
> with open('file') as f:
>         a: str = f.read()
> ```
> 
> If you'd like I can issue a PR for that too.

That would be great

hpapaxen:
> > @soupytwist I also have a fix for recognizing type annotations in python 3. i.e.
> > ```
> > with open('file') as f:
> >         a: str = f.read()
> > ```
> > If you'd like I can issue a PR for that too.
> 
> That would be great

PR is #268 

Cheers.

PR #268, opened 2019-03-20
Assign annotation

hpapaxen:


hpapaxen:
@soupytwist You think we can bump up a minor version after this merge?

PR #269, opened 2019-03-22
Fix return value of __enter__ for builtin File

hpapaxen:
When __enter__ is called within a 'with' expression, wrap the
return value with a pyobject instead of returning 'File' directly.
This will keep us consistent with existing functionality


hpapaxen:
@soupytwist Can you review?

PR #280, opened 2019-12-18
Replace all deprecated assertEquals with assertEqual (and stop supporting 2.6)

mcepl:


mcepl:
@joystein Could I ask for the review here?

joystein:
@mcepl, did you intend to address @soupytwist  in your comment above? I can attempt to give a review, but I'm not all that familiar with the review process on github.

I looked at this yesterday, actually. I also tried to run the tests with pytest and saw all the warnings. I guess the commit that replaces the asserts is a find and replace? Or did you use rope itself somehow? I ran the tests suite for master and your branch and they both came out with 192 tests passed, so I guess that part is ok.

The part about raw strings I don't know much about.

I guess that if you intend to transition to pytest, that you could then remove the test command in setup.py as well. And you might want to add an `extras_require` section in setup.py for developer convenience (so that dev dependencies can be pulled in by `pip install -e .[dev]`):

```python
extras_require={
        "dev": [
            "pytest",
        ]
    },
```

joystein:
Nice changes! Would it make sense to add a sentence or two about setup of development environment and running of tests to the top-level README?

mcepl:
Anything else?

joystein:
Looks good to me!

PR #285, opened 2020-01-14
Make rope compatible with Python 3.8

mcepl:
Fixes for #284

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
   - "3.5"
   - "3.6"
   - "3.7"
+  - "3.8"
   - "pypy3"
 matrix:
   allow_failures:
@@ -16,4 +17,4 @@ install:
     - pip install pytest
 notifications:
     irc: "chat.freenode.net#rope"
-script: pytest -v -k 'not (test_hint_parametrized_iterable or test_hint_parametrized_iterator)'
+script: pytest -v
--- a/rope/base/change.py
+++ b/rope/base/change.py
@@ -116,10 +116,10 @@ class ChangeSet(Change):
 
 
 def _handle_job_set(function):
-    """A decorator for handling `taskhandle.JobSet`\s
+    """A decorator for handling `taskhandle.JobSet`
 
-    A decorator for handling `taskhandle.JobSet`\s for `do` and `undo`
-    methods of `Change`\s.
+    A decorator for handling `taskhandle.JobSet` for `do` and `undo`
+    methods of `Change`.
     """
     def call(self, job_set=taskhandle.NullJobSet()):
         job_set.started_job(str(self))
--- a/rope/base/evaluate.py
+++ b/rope/base/evaluate.py
@@ -196,6 +196,14 @@ class StatementEvaluator(object):
         type_name = type(node.n).__name__
         self.result = self._get_builtin_name(type_name)
 
+    def _Constant(self, node):
+        type_name = type(node.n).__name__
+        try:
+            self.result = self._get_builtin_name(type_name)
+        except exceptions.AttributeNotFoundError:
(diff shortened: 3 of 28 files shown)


tgy:
Hey, any progress on this @mcepl? need a hand?

mcepl:
> Hey, any progress on this @mcepl? need a hand?

Too much overwhelmed with other stuff. Any PRs would be of course graciously considered.

last-partizan:
I tried to dig this issue, and there is two problems:

1. Test case in #284 is bad. It checks one dynamic result against another, and under 3.8 both are invalid just in different ways. So i added expected results and now it looks better and works as expected.

```.diff
diff --git a/ropetest/advanced_oi_test.py b/ropetest/advanced_oi_test.py
index 02a989b2..223babb2 100644
--- a/ropetest/advanced_oi_test.py
+++ b/ropetest/advanced_oi_test.py
@@ -238,15 +238,25 @@ class DynamicOITest(unittest.TestCase):
         def complex_to_textual(pyobject):
             return to_textual.transform(
                 to_pyobject.transform(to_textual.transform(pyobject)))
-        for name in ('C', 'f', 'a_var', 'a_list', 'a_str', 'a_file'):
-            var = pymod[name].get_object()
-            self.assertEqual(to_textual.transform(var),
-                              complex_to_textual(var))
-        self.assertEqual(to_textual.transform(pymod),
-                          complex_to_textual(pymod))
-        enumerate_func = rope.base.builtins.builtins['enumerate'].get_object()
-        self.assertEqual(to_textual.transform(enumerate_func),
-                          complex_to_textual(enumerate_func))
+
+        test_variables = [
+            ('C', ('defined', 'mod.py', 'C')),
+            ('f', ('defined', 'mod.py', 'f')),
+            ('a_var', ('instance', ('defined', 'mod.py', 'C'))),
+            ('a_list',
+             ('builtin', 'list', ('instance', ('defined', 'mod.py', 'C')))),
+            ('a_str', ('builtin', 'str')),
+            ('a_file', ('builtin', 'file')),
+        ]
+        test_cases = [(pymod[v].get_object(), r) for v, r in test_variables]
+        test_cases += [
+            (pymod, ('defined', 'mod.py')),
+            (rope.base.builtins.builtins['enumerate'].get_object(),
+             ('builtin', 'function', 'enumerate'))
+        ]
+        for var, result in test_cases:
+            self.assertEqual(to_textual.transform(var), result)
+            self.assertEqual(complex_to_textual(var), result)
 
     def test_arguments_with_keywords(self):
         mod = testutils.create_module(self.project, 'mod')
```

2. Second problem: there is ast.Constant, which is unsupported by `StatementEvaluator`, it looks like it can be fixed this way:
```.diff
diff --git a/rope/base/evaluate.py b/rope/base/evaluate.py
index ce38bd02..62eeed72 100644
--- a/rope/base/evaluate.py
+++ b/rope/base/evaluate.py
@@ -196,6 +196,14 @@ class StatementEvaluator(object):
         type_name = type(node.n).__name__
         self.result = self._get_builtin_name(type_name)
 
+    def _Constant(self, node):
+        type_name = type(node.n).__name__
+        try:
+            self.result = self._get_builtin_name(type_name)
+        except exceptions.AttributeNotFoundError:
+            # XXX: Right way to fix this is to add missing NoneType to builtins?
+            pass
+
     def _get_builtin_name(self, type_name):
         pytype = rope.base.builtins.builtins[type_name].get_object()
         return rope.base.pynames.UnboundName(
```

Code for `_Constant` is almost the same as for `_Num`, so it can be refactored into using single function if needed. My concern is this: is this safe to just catch excetion here? or maybe better way to add NoneType to builtin types?

Issue #288, opened 2020-02-05
Rope renames variable incorrectly


hbusul:
I've come across with this issue while using Rope as a library.

Python version: 3.7.6
OS: Ubuntu 18.04
Rope Version: 0.16.0

I also noticed this bug does not occur on Rope version 0.14.0

## How to reproduce the error
1. Create the following file structure
```
app
├── run.py
└── my_app
|   ├── __init__.py
|   └── apis
|      └── __init__.py
|      └── my_namespace.py
```

```
#app/run.py
from my_app import app

if __name__ == '__main__':
    app.run(host='0.0.0.0')

#app/my_app/__init__.py
from flask import Flask
from flask_httpauth import HTTPBasicAuth, HTTPTokenAuth, MultiAuth

app = Flask(__name__)
auth = MultiAuth(HTTPBasicAuth(), HTTPTokenAuth('Bearer'))

from my_app.apis import api  # noqa
api.init_app(app)

#app/my_app/apis/__init__.py
from flask_restplus import Api
from apis.my_namespace import api as my_api


api = Api()
api.add_namespace(my_api)


#app/my_app/apis/my_namespace.py
from flask_restplus import Namespace, Resource
from my_app import auth

api = Namespace('somenamespace')


class MyClassA(Resource):
    @auth.login_required
    def delete(self):
        my_var = f"some value"


class MyClassB(Resource):
    @auth.login_required
    def get(self):
        pass


api.add_resource(MyClassA, '/class-a')
api.add_resource(MyClassB, "/class-b")
```

2. Create the script for renaming `auth` variable in `app/my_app/__init__.py`.
This script should be at the same level with app folder. Name it `renamer.py`.
```
from rope.base import libutils, project
from rope.refactor.rename import Rename


def main():
    my_project = project.Project("app")
    file = "app/my_app/__init__.py"
    file_as_res = libutils.path_to_resource(my_project, file)
    declaration = 115
    change_set = Rename(my_project, file_as_res,
                        declaration).get_changes("f162b77e780b")
    my_project.do(change_set)
    my_project.close()


if __name__ == "__main__":
    main()
```

3. Call that script `python renamer.py`

## Expected Result
In the `app/my_app/apis/my_namespace.py` it should look like:
```
from flask_restplus import Namespace, Resource
from my_app import f162b77e780b

api = Namespace('somenamespace')


class MyClassA(Resource):
    @f162b77e780b.login_required
    def delete(self):
        my_var = f"some value"


[...]

PR #289, opened 2020-02-13
get_string_pattern_with_prefix short str pattern modified

hbusul:
These are the changes to fix the bug described at [issue 288](https://github.com/python-rope/rope/issues/288) . It is passing all the tests and also solving the issue. Could you take a look at it please?

### DIFF
--- a/ropetest/codeanalyzetest.py
+++ b/ropetest/codeanalyzetest.py
@@ -347,10 +347,6 @@ class WordRangeFinderTest(unittest.TestCase):
         finder = worder.Worder(code)
         self.assertEqual(finder.get_word_range(45), (45, 49))
 
-    def test_get_word_range_with_fstring2(self):
-        code = "auth = 8\nmy_var = f'some value {auth}'\nprint(auth)\nother_val = 'some other'"
-        finder = worder.Worder(code)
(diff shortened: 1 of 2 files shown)


ropetest/codeanalyzetest.py
```diff
@@ -342,6 +342,15 @@ def test_is_on_function_keyword(self):
         finder = worder.Worder(code)
         self.assertTrue(finder.is_on_function_call_keyword(len(code) - 1))
 
+    def test_get_word_range_with_fstring(self):
+        code = 'auth = 8\nmy_var = f"some value {auth}"\nprint(auth)\nother_val = "some other"'
+        finder = worder.Worder(code)
+        self.assertEqual(finder.get_word_range(45), (45, 49))
+
+    def test_get_word_range_with_fstring2(self):
```
soupytwist: nit: This test is a bit unnecessary.

soupytwist:
Looks great and makes sense. It's not a huge deal, but if you can, please remove the redundant test here to keep down on clutter.

Will merge soon regardless. Thanks for the fix!

hbusul:
For sure, removed the second test.

soupytwist:
Thanks again!

PR #297, opened 2020-05-15
Fix `Worder.get_primary_at` cutting off names that start like keywords.

### LINKED ISSUE #296: `Worder.get_primary_at` cuts off names that start like keywords.
Calling `Worder.get_primary_at` on code with names that start with Python keywords (but are otherwise valid names) gives bad results when given an offset exactly one character before the end of the part of the name that looks like a keyword.

For example:

```python
>>> from rope.base.worder import Worder
>>> Worder("is_keyword = False\n").get_primary_at(1)
's_keyword'
>>> Worder("global_valid_name = 10\n").get_primary_at(5)
'l_valid_name'
```

which is obviously not right.

This causes a number of strange issues with other `rope` tools that use `Worder`.

thekrampus:
E.g. names like `is_whatever` or `global_whatever`.

- Adds test for finding names that start with keywords.
- Fixes `Worder.get_primary_at` cutting off names that start with keywords.

Closes #296

### DIFF
--- a/rope/base/worder.py
+++ b/rope/base/worder.py
@@ -209,7 +209,8 @@ class _RealFinder(object):
         if offset >= 0 and (self.code[offset] in '"\'})]' or
                             self._is_id_char(offset)):
             atom_start = self._find_atom_start(offset)
-            if not keyword.iskeyword(self.code[atom_start:offset + 1]):
+            if not keyword.iskeyword(self.code[atom_start:offset + 1]) or \
+               (offset + 2 < len(self.code) and self._is_id_char(offset + 2)):
                 return atom_start
         return last_atom
 
--- a/ropetest/codeanalyzetest.py
+++ b/ropetest/codeanalyzetest.py
@@ -111,6 +111,11 @@ class WordRangeFinderTest(unittest.TestCase):
         result = self._find_primary(code, code.index('attr') + 1)
         self.assertEqual('a_var.\nattr', result)
 
+    def test_word_finder_on_primary_like_keyword(self):
+        code = 'is_keyword = False\n'
+        result = self._find_primary(code, 1)
+        self.assertEqual('is_keyword', result)
+
     def test_strings(self):
         code = '"a string".split()'
         self.assertEqual('"a string".split', self._find_primary(code, 14))


thekrampus:
I'm not totally sure how to set up a development environment... When I run the tests with `pytest ropetest/**.py` a whole load of tests fail on `master`, i.e. **not including this patch.** I'm using python 3.8, setting up my environment with `pip install -e .[dev]`

In any case this patch does not cause any more tests to fail then were already failing.

mcepl:
Yes, we had broken `master` until now. I am sorry for that.

@lieryan What do you think about this?

mcepl:
> @mcepl On the branch with all 3.8 test fixed, the new test failed before this fix is applied and then passed after the fix is applied.

I don’t understand, when rebased over 33d5e667, the testsuite passes (with Python 3.8) “1723 passed, 14 skipped, 1631 warnings in 10.58s”.

lieryan:
Ah sorry, I should've written more clearly, I started writing the first paragraph of my comment above before I discovered the bug described in the second paragraphs onwards. The "this fix" in my first paragraph referred to thekrampus' fix, not my proposed modifications to thekrampus' fix; while the "this test case" and "this change" in my second and third paragraph was referring to my proposed modifications.

I was red-green testing thekrampus' patch to verify whether the new test he was adding was reproducing the bug in all the tox environments or just Py3.8, so I cherry pickes only the new test case without his_find_primary_without_dot_start fix so I can see what the original code was doing. His test did reproduce the issue in all versions, and applying his fix passed all tests in all versions. So in short, all is well on that side.

The second and third paragraph is a test case demonstrating a case that thekrampus' patch didn't handle correctly.

mcepl:
@lieryan So, to summarize: you think this is a valid pull request, deals with the real issue, but it doesn't cover all possible tests (especially the one you suggest)?

lieryan:
Yes, that summary sounds about right.

> So, you would like to see this included mcepl@dae8b5a ?

Yes.

mcepl:
Merged as 35846c67. Additional commit by @lieryan is dae8b5a6.

mcepl:
Thank you.

PR #306, opened 2020-09-19
Fix patchedast for Python 3.8

lieryan:
Fix most of the errors in patchedast due to changes in the ast module in Python 3.8. 

In the future, patchedast probably could be simplified a lot removing most of the regex parsing it currently does and making use of the new `end_lineno` and `end_col_offset` attribute, but since we probably still need to support older python, it may take a while until rope can fully take advantage of those attributes.

This branch also fixes some bugs with bytestring literal syntax (`b""`).

Before:

```
(rope-py3.8)$ pytest
[snip]
139 failed, 1573 passed, 14 skipped, 1807 warnings
```

After:

```
(rope-py3.8)$ pytest
[snip]
3 failed, 1711 passed, 14 skipped, 1621 warnings
$ tox
[snip]
  py27: commands succeeded
  py34: commands succeeded
  py35: commands succeeded
  py36: commands succeeded
  py37: commands succeeded
ERROR:   py38: commands failed
```

There are still some tests failing in Python 3.8 in codeassist module, but I think they should be fixed as a separate PR.

PR #307, opened 2020-10-04
Fix get definition location for Python 3.8

lieryan:
This PR changes `get_definition_location()` to point to the line where the class/function is defined rather than pointing to the decorator. This makes older versions match Python 3.8's more correct behaviour.

Note that this is a backwards incompatible change if someone depends on `get_definition_location()` pointing to the decorator.

Note: this PR also contains commits from https://github.com/python-rope/rope/pull/306, you most likely would want to review the diffs just from this branch to that one rather than seeing both changes simultaneously.

### DIFF
--- a/rope/base/pynames.py
+++ b/rope/base/pynames.py
@@ -21,7 +21,8 @@ class DefinedName(PyName):
         return self.pyobject
 
     def get_definition_location(self):
-        return (self.pyobject.get_module(), self.pyobject.get_ast().lineno)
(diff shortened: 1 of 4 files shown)


mcepl:
I see af33ddbb4f4bfe57a7e613691959891dc6824147 as the only commit on the top `master` (after merging all other PRs), is it correct? Could you please rebase?

lieryan:
I've rebased the branch.

Yes, this is a single-commit PR.

Issue #248, opened 2018-07-04
get_definition_location() goes to wrong instance


shaolo1:
```
def foo(self):
    a = 0
    p = 'foo'''  # The extra single quote pair at the end is what breaks things!..without them it works as expected


def bar(self):
    a = 1
    a += 1  # using get_definition_location() on the a in this line will report line 2 vs 6
```

shaolo1:
It is an odd case, but p = 'foo''bar' is valid python3 making p== 'foobar'.  This actually works with get_definition_location(), so it looks like its just the special case where the second string is empty that breaks things.

soupytwist:
Very weird... sorry for the non-response, I don't have a good idea what the fix for this will entail. If you have the time to look into it, it would be appreciated, otherwise I will revisit when I have more time.

shaolo1:
I "fixed" the extra quotes in my code, so get_definition_location() is working again for me, but I wanted to make sure the issue was logged in case someone else hit it.

PR #309, opened 2020-10-11
#248 fix false triple quoted string

lieryan:
Fix the linked issue #248. 

This issue is because the three single quotes here: `p = 'foo'''` is misrecognised as a triple-quoted string, which is not consistent with how Python interprets it as an opening and closing of single-quoted string.


soupytwist (approved):

soupytwist:
Thanks for rooting this out and sending a patch!

lieryan:
Thanks for reviewing and merging this.

Issue #249, opened 2018-07-06
Rope fails to rename variable identifier within a generator expression 


d3r3kk:
Issue found with renaming a variable within a generator expression. Exception thrown is:

**RefactoringError: Rename refactoring should be performed on resolvable python identifiers.**

**Python version:** 3.6.6
**OS**: Ubuntu 18.04
**Rope Version**: 0.10.7

## Reproduction Steps
1. Create a folder `project`
2. `cd project`
3. Create a python file called issue.py and copy the following code into it:
````
def do_stuff():
    my_list = [counter + 10 for counter in range(0, 10)]
    print(my_list)

do_stuff()
````
4. Create another python file named renamer.py and copy the following code into it:
````import rope
from rope.base import libutils
from rope.refactor.rename import Rename
import rope.base.project

proj = rope.base.project.Project(
    './',
    save_history=False,
    indent_size=4)

res = libutils.path_to_resource(proj, 'issue.py')

if res.exists:
    renamer = Rename(proj, res, 24)
    changes = renamer.get_changes('valid_count', resources=res)
    for c in changes.changes:
        print(c.new_contents)

````
5. Run the renamer script via the command line `> python renamer.py`

## Expected result
A printout of the `issue.py` file's contents, with the 2 instances of the 'counter` variable renamed to `valid_count`.

## Observed result
Thrown exception:
````
Traceback (most recent call last):
  File "renamer.py", line 14, in <module>
    renamer = Rename(proj, res, 31)
  File "/home/dekeeler/.local/lib/python3.6/site-packages/rope/refactor/rename.py", line 28, in __init__
    'Rename refactoring should be performed'
rope.base.exceptions.RefactoringError: Rename refactoring should be performed on resolvable python identifiers.
````

phoenixsampras:
Hello and help, any workaround for this issue?

soupytwist:
Sorry, I haven't had much time to look into these issues further. If anyone else here is able to pick it up, that would be appreciated.

wheelsandmetal:
Any news on this?

mosesontheweb:
I still encounter this behaviour in v0.16.0. Anyone got tipps? Using it within VSCode.

rogedelgado:
Hello,

I can confirm the issue.I got the error in vim using LanguageCliente-neovim as client. The rope version that I'm using is v0.16.0.

Any advice for this?

ArkadyWey:
Hi, 

I am also still having this issue.

[...]

PR #318, opened 2020-11-08
Implement renaming of walrus operator and loop variables in comprehensions

lieryan:
This PR implements renaming variables assigned by inline assignment expression (a.k.a. walrus operator/`:=`) and the looping variables of list/set/dict/generator comprehension.

This also solves point 3 from [my comment on #249](https://github.com/python-rope/rope/issues/249#issuecomment-718248829). The variable scoping for list comprehension does not follow Python 3 semantic, but as mentioned there, fixing the scopes likely will require a significant rewrite of the scope finding mechanism.


mcepl:
LGTM. Thank you.

PR #322, opened 2020-12-06
Fix #114. Extract augmented assignment

### LINKED ISSUE #114: Problem in ExtractMethod when incrementing a variable.
Let's say I have the following code:

``` python
i = 1
i += 1
print(i)
```

It prints two.

When extracting the second line, I would expect:

``` python
i = 1

def extracted(i)
    return i + 1

i = extracted(i)
print(i)
```

I get instead:

``` python
i = 1

def test():
    i += 1
    return i

i = test()
print(i)
```

which clearly does not work.

I know nothing about the complexity of managing assignment operators, but if it is not supported at the moment, it would be better to block the refactoring instead of creating a wrong one.

Thanks for your help.

lieryan:
Fix issue in #114 where extracting code that writes to global variables or contains augmented assignments produces refactored code that doesn't behave the same as the original code.

### DIFF
--- a/rope/base/evaluate.py
+++ b/rope/base/evaluate.py
@@ -235,7 +235,7 @@ class StatementEvaluator(object):
         keys = None
         values = None
         if node.keys and node.keys[0]:
-            keys, values = next(filter(itemgetter(0), zip(node.keys, node.values)), (None, None))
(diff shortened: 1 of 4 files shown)


mcepl:
Thank you.

PR #325, opened 2020-12-26
Check for Nonetype Node

shanechin:
I'm still trying to figure out what is going on here - so apologize for being light on details.

I have been trying to do a file rename refactor, which fails with:
```
>>> AttributeError: 'NoneType' object has no attribute '_fields'
```

Trying to see what the reproduction scenario is, but I *think* it is handling client code that looks something like this:

Operation: rename c.py -> z.py

```
# filename a/e.py

def foo():
       - from a.b import c 
       + from a.b import z
       z.foo()
```

Taking a look at 
https://github.com/python-rope/rope/blob/65080c52d5e1d6c196417b7c877af89ff44f5278/rope/base/evaluate.py#L289-L291

It seems I have a function scope with no node.

### DIFF
--- a/rope/base/ast.py
+++ b/rope/base/ast.py
@@ -46,7 +46,7 @@ def get_child_nodes(node):
     if isinstance(node, ast.Module):
         return node.body
     result = []
-    if node._fields is not None:
+    if node and node._fields is not None:
         for name in node._fields:
             child = getattr(node, name)
             if isinstance(child, list):


mcepl:
Is this PR still in progress or do you ask me to review it and merge it?

shanechin:
@mcepl - I don't have context on the what this might break - I would like for you to review / merge if you think the "fix" here is sane. 

It seems to me like we shouldn't be getting this low in the call stack - as in some higher level assertion/constraint is wrong, but I would need to understand the codebase much more to fix that. 

This change seems to keep the type signature the same - and do something I believe is sane.

mcepl:
> @mcepl - I don't have context on the what this might break - I would like for you to review / merge if you think the "fix" here is sane.

The honest truth is that I have no clue about most of rope internals. Most of my work in the last couple of years was to accepting pull requests and making sure things look at least from distance. Given there is nobody who would be working on rope more than me, I guess at least this kind of involvement if better than letting this project completely died, which it was doing before I started with it.

> It seems to me like we shouldn't be getting this low in the call stack - as in some higher level assertion/constraint is wrong, but I would need to understand the codebase much more to fix that.
> 
> This change seems to keep the type signature the same - and does something I believe is sane.

Well, changing of the type of the function parameter from `something` to `Optional[something]` (using PEP-484 notation) is still changing the signature of the function, so it makes me a little bit worried.

I think, it could be productive to investigate which calls make the function `get_child_nodes()` to get empty `node`, and to decide whether it is a legitimate situation or not.

Also, some tests for the calling functions making them to have the `node` parameter `None` (if it is a legitimate situation) would be helpful.

lieryan:
I was not able to reproduce this bug given the supposed scenario:

```
$ git show 661510f83719aee68f26212d595af726e90a0ba5   ## 0:40:01
commit 661510f83719aee68f26212d595af726e90a0ba5
Author: Lie Ryan <lie.1296@gmail.com>
Date:   Thu Sep 9 00:39:07 2021 +1000

    original

diff --git a/a/b/c.py b/a/b/c.py
new file mode 100644
index 0000000..82089cf
--- /dev/null
+++ b/a/b/c.py
@@ -0,0 +1,2 @@
+def foo():
+    print('in foo')
diff --git a/a/b/z.py b/a/b/z.py
new file mode 100644
index 0000000..e69de29
diff --git a/a/e.py b/a/e.py
new file mode 100644
index 0000000..88a6a89
--- /dev/null
+++ b/a/e.py
@@ -0,0 +1,5 @@
+def foo():
+    from a.b import c
+    c.foo()
+
+foo()
```

When I use rename refactor on `c.py`, it renames the module to `z.py` correctly as it should:

```
$ git show   ## 0:41:10
commit d542fa0fa3ab5e32c941deaaf2709668466ee498 (HEAD -> master)
Author: Lie Ryan <lie.1296@gmail.com>
Date:   Thu Sep 9 00:39:17 2021 +1000

    move a/b/c.py to a/b/z.py using rename

diff --git a/a/b/c.py b/a/b/c.py
deleted file mode 100644
index 82089cf..0000000
--- a/a/b/c.py
+++ /dev/null
@@ -1,2 +0,0 @@
-def foo():
-    print('in foo')
diff --git a/a/b/z.py b/a/b/z.py
index e69de29..82089cf 100644
--- a/a/b/z.py
+++ b/a/b/z.py
@@ -0,0 +1,2 @@
+def foo():
+    print('in foo')
diff --git a/a/e.py b/a/e.py
index 88a6a89..0deadb4 100644
--- a/a/e.py
+++ b/a/e.py
@@ -1,5 +1,5 @@
 def foo():
-    from a.b import c
-    c.foo()
+    from a.b import z
+    z.foo()
 
 foo()
```

lieryan:
Hi @shanechin, thanks for taking the time to write this bug and PR. As you suspect, the `node` being None likely should never have reached this function in the first place, so there is likely something higher in the call stack responsible for the issue. But unfortunately there is not enough detail here to figure out what they were in this scenario. 

I'm going to close this ticket for now, but if you encounter this issue again, please provide the following information:

1. Rope version
2. IDE/Editor name and version
3. What editor plugin you used for rope and their version
4. Python version that Rope is running at
5. A minimal, complete, and reproducible example describing what you're attempting to do

PR #332, opened 2021-02-28
create_generate with goal_resource param

climbus:
I'm working on option generating class in different file in ropevim. RopeVim uses "create_generate" function, it doesn't provide option to set "goal_resource".

I added this to rope. It will be nice if this can be merged. Thanks ;)


mcepl:
Go for it! Thanks.

Issue #324, opened 2020-12-13
Long time taking to refactor


XtremeGood:
I don't know but it takes too long to refactor on my win 10 , 64 bit, i7 7th gen , 16 GB RAM laptop.

![image](https://user-images.githubusercontent.com/40588378/102024714-acaa0600-3db9-11eb-8745-5f50bc619410.png)

mcepl:
If you are using VS Code plugin, I am not completely persuaded they actually use this version of rope, or they bundled some of older (and probably even patched) version.

XtremeGood:
Well I ‘m using vscode but it did a pip install in the virtualenv. Sent from Mail for Windows 10 From: Matěj CeplSent: 14 December 2020 06:15To: python-rope/ropeCc: Vishesh Mangla; AuthorSubject: Re: [python-rope/rope] Long time taking to refactor (#324) If you are using VS Code plugin, I am not completely persuaded they actually use this version of rope, or they bundled some of older (and probably even patched) version.—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.

zevaverbach:
Chiming in here with a +1. I'm on a modern Mac (2019 16" MBP) using VSCode, and the "rename" refactor takes ages to run.

mcepl:
@zevaverbach , @XtremeGood could you both quantify “ages” and “too long”, please? Are we talking about seconds, minutes?

zevaverbach:
@mcepl many seconds, upwards of 20-30.

mcepl:
@zevaverbach And what size of the project? How many files, how many megabytes?

XtremeGood:
just a single file of 30-40 lines

lieryan:
> just a single file of 30-40 lines

That doesn't sound like a performance issue with rope, but rather how you development environment is set up. I've used rename refactor on really big Python projects, like 200kloc projects, and rename never takes more than two seconds to complete.

I am not familiar with VScode, but I noticed you might have a venv folder on your directory. Try moving your virtualenv folder to somewhere outside the rope project folder and see if you're still having an issue.

mcepl:
I cannot reproduce here. Renaming with neovim+python-language-server (which uses rope for refactoring) takes a second or two, but nothing too crazy (and LSP itself adds substantial overhead).

euseanwoon2016:
[...]

PR #358, opened 2021-08-01
Modifies default_config.py

0x1e02:
In the spirit of #324  I could verify that long waiting times are a result of not ignoring `venv` or `.venv` (pick your poison). As it is best practice to use venv I think they should be ignored by default. Also this cost me way to much headache to realize and I can imagine  that I am not alone in that regard.

Greetings,
Leon


mcepl:
Thank you.

PR #361, opened 2021-08-23
Discuss: should we move off the Google Groups developer mailing list to Github?

lieryan:
As of the time of posting, the developer's Google Group seems to be dead. Unless someone has the know-how and ability to restore the Google Groups mailing list and take ownership to manage it, then it would make sense to just abandon it and do all dev discussion in Github.

Should we just move completely to Github? Maybe enable Github Discussions for the repository?

This Pull Request does not constitute a decision to abandon the mailing list. If we decide to continue with the mailing list, then this pull request should be closed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -40,10 +40,9 @@ you open a project).
 Bug Reports
 ===========
 
-Send your bug reports and feature requests to `rope-dev (at)
-googlegroups.com`_.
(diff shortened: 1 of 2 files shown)


mcepl (approved):
LGTM

lieryan:
I hadn't heard any opposing opinions, so I've proceeded to set up Github Discussions in the repository.

PR #365, opened 2021-09-04
Support extract method of expressions with inline assignment (walrus operator)

### LINKED ISSUE #357: Rope does not support extract method on expressions with Inline Assignment (walrus operator)
I get a syntax error on something like this:

```
if match := cls._REGEX_PATTERN.match(field.name):
    pass
```

lieryan:
Fix #357.


mcepl (approved):
LGTM

PR #126, opened 2015-09-14
WIP: Not-quite-working exec patch

ceridwen:
I followed the pattern in _excepthandler as much as possible, but the tests are still failing, I suspect because the exec function has parentheses instead of white space, and I don't know how to handle the case where symbols are optional and can't find any analogous cases in the rest of the code.  Can you point me to what I need to add?

Fixes #117

PR #372, opened 2021-09-08
Fix MismatchedTokenError caused by use of exec with parentheses in Python 2

### LINKED ISSUE #117: MismatchedTokenError caused by use of exec() function
The problem is that `ast` compiles both the exec statement and exec function into the same AST nodes, while patchedast.py assumes the `exec` statement.

``` python
>>> import ast
>>> ast.Exec
<class '_ast.Exec'>
>>> ast.Exec._fields
('body', 'globals', 'locals')
>>> ast.dump(ast.parse('exec "def f(): pass" in {}'))
"Module(body=[Exec(body=Str(s='def f(): pass'), globals=Dict(keys=[], values=[]), locals=None)])"
>>> ast.dump(ast.parse('exec("def f(): pass", {})'))
"Module(body=[Exec(body=Str(s='def f(): pass'), globals=Dict(keys=[], values=[]), locals=None)])"
```

From `patchedast.py`, lines 335-342.

``` python

    def _Exec(self, node):
        children = []
        children.extend(['exec', node.body])
        if node.globals:
            children.extend(['in', node.globals])
        if node.locals:
            children.extend([',', node.locals])
        self._handle(node, children)
```

I'm willing to write a patch and pull request, but what's the right way to rewrite _Exec?

lieryan:
Original patch was created by @ceridwen, thank you for your contribution!

The original patch looks good, this PR just made minor cleanup and added some tests.

Fix #117.
Close #126.

### DIFF
--- a/rope/refactor/patchedast.py
+++ b/rope/refactor/patchedast.py
@@ -460,25 +460,6 @@ class _PatchingASTWalker(object):
         children.append(self.exec_close_paren_or_space)
         self._handle(node, children)
 
-    # def _Exec(self, node):
-    #     children = []
-    #     children.extend(['exec', '(', node.body])
-    #     if node.globals:
-    #         children.extend([',', node.globals])
-    #     if node.locals:
-    #         children.extend([',', node.locals])
-    #     children.append(')')
-    #     self._handle(node, children)
-
-    # def _Exec(self, node):
-    #     children = []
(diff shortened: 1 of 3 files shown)


lieryan:
Currently, refactoring code containing `exec` has the limitation that code scope isn't currently handled properly. I've split off that issue into a separate ticket: #373.

PR #380, opened 2021-09-12
Fix variables that are returned and/or turned into argument when extracting method in a loop

### LINKED ISSUE #112: Unexpected behavior when extracting loop body
Program versions:
- rope==0.10.2
- ropemode==0.2
- ropevim==0.5.0

Let's say I have the following function:

``` python
def my_func():
    i = 0
    for dummy in range(10):
        i += 1
        print i
```

When I select the i += 1 for method extraction, I get:

``` python
def my_func():
    i = 0
    for dummy in range(10):
        i = increment(i)
        print i

def increment(i):
    i += 1
    return i
```

which is exactly what I would expect.

But if I extractmethod all of the loop body instead, I get:

``` python
def my_func():
    i = 0
    for dummy in range(10):
        increment_and_print(i)

def increment_and_print(i):
    i += 1
    print i
```

which clearly gives me a different output.

It seems that variables that are used inside the body are not detected as necessary return values, which may lead to erratic behavior and bugs (which will be hard to track down).

If this is known because too difficult to implement, then I think it would be really nice to have a list of the Rope known limitations, so as to know its weaknesses.

Thanks.

lieryan:
Fix #112.

- [x] Create ticket: https://github.com/python-rope/rope/issues/393

### DIFF
--- a/rope/refactor/extract.py
+++ b/rope/refactor/extract.py
@@ -1,10 +1,11 @@
 import re
+from contextlib import contextmanager
 
-from rope.base.utils.datastructures import OrderedSet
 from rope.base import ast, codeanalyze
 from rope.base.change import ChangeSet, ChangeContents
 from rope.base.exceptions import RefactoringError
 from rope.base.utils import pycompat
+from rope.base.utils.datastructures import OrderedSet
 from rope.refactor import (sourceutils, similarfinder,
                            patchedast, suites, usefunction)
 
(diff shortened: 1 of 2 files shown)


lieryan:
One maybe-bug with this refactoring:

```
def my_func():
    for dummy in range(10):
        i += 1
```

Current extract refactoring implementation produces:

```
def my_func():
    for dummy in range(10):
        i = aaa()

def aaa():
    i += 1
    return i
```

Arguably, the original code was non-sense to begin with, as `i` was undefined. When run, that code would've produced a `NameError`, so strictly speaking, this refactoring is garbage-in, garbage-out.

If `i` was assigned before, such as the below code, the current implementation would have produced correct refactoring:

```
def my_func():
    i = 0
    for dummy in range(10):
        i += 1
```

Maybe rope could catch this and refuse to refactor such code.

PR #389, opened 2021-09-13
Implement `async def`, `async for`, and `await` syntax for method extraction

### LINKED ISSUE #375: Implement extract refactoring inside `async def`
```
async def my_func(my_list):
    for x in my_list:
        var = x + 1
    return var
```

Expected

```
async def my_func(my_list):
    for x in my_list:
        var = new_func(x)
    return var

def new_func(x):
    var = x + 1
    return var
```

Actual: 

> [Pymode]: error: Unhandled exception in Pymode: 'For' object has no attribute 'region'

lieryan:
Fix #375 

Todo:

- [x] fix version specifier
- [x] https://github.com/python-rope/rope/pull/389#issuecomment-918436011
- [x] https://github.com/python-rope/rope/pull/389#issuecomment-918439095

### DIFF
--- a/rope/refactor/extract.py
+++ b/rope/refactor/extract.py
@@ -754,6 +754,9 @@ class _FunctionInformationCollector(object):
             for name in visitor.read - visitor.written:
                 self._read_variable(name, node.lineno)
 
+    def _AsyncFunctionDef(self, node):
+        self._FunctionDef(node)
+
     def _Name(self, node):
         if isinstance(node.ctx, (ast.Store, ast.AugStore)):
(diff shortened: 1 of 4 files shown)


lieryan:
Currently failing cases:

```
            async def my_func(my_list):
                async for x in my_list:
                    var = x + 1
                return var
```

Extracting line 2-3 should produce:

```
            async def my_func(my_list):
                var = new_func()
                return var

            async def new_func():
                async for x in my_list:
                    var = x + 1
                return var
```

PR #391, opened 2021-09-16
Extract with assignment issue 376

### LINKED ISSUE #376: Extract method only with instance variable change assigment
Given class:

```
class A(object):
    def func(self):
        self.var_a = 1
        var_bb = self.var_a
```

When extracting method with selected instance variable  `var_bb = [self.var_a]` and selected similar to True

Then changing var_a assignment:
```
class A(object):
    def func(self):
        self.new_func() = 1
        ^^^^^^^^^^^^^^^
        var_bb = self.new_func()

    def new_func(self):
        return self.var_a
```

Should be:

```
class A(object):
    def func(self):
        self.var_a = 1
        var_bb = self.new_func()

    def new_func(self):
        return self.var_a

```

climbus:
Fix #376

Fix bug with replacing right side of assignment. I don't know ast well. If you know any cases which may not work, please send me an example.


lieryan (approved):

lieryan (approved):
LGTM

Issue #104, opened 2015-01-11
Failing parse on nested with-statements.


Ferada:
I've uploaded <del>[this commit](https://github.com/Ferada/rope/commit/6f7cacc9f261430740086748ae644ccd942ed578) with two failing tests</del><ins>[this commit](https://github.com/Ferada/rope/commit/f24503b255ad033572f7e4da13686375996d0c02) with one failing test case which</ins><del>, one just fails to refactor correctly, the other</del> raises an exception because<del>(?)</del> the source code doesn't exactly match the AST, i.e.:

``` python
    with open("test") as file1, open("test") as file2:
        print(file1, file2)
```

fails to parse; there is a second (nested) `with` node in the AST, but in `_Source.consume`/`patchedast.py:643` the `self.offset` is at the end of the line and couldn't match a second, non-existent `with` anyway.

<del>
The other, simpler case is a regular nested `with`, i.e. extracting a method from the following code (the `print` line):

````python
    with open("test") as file1:
        with open("test") as file2:
            print(file1, file2)
````

will refactor (and miss the two variables) to:

````python
    def extracted():
        print(file1, file2)

    with open("test") as file1:
        with open("test") as file2:
            extracted()
````
</del>

I'm pretty sure I can figure out the second case, but for the first I'm at a loss how to start fixing it. Any pointers?

Edit: I've changed the one test case to reflect the working behaviour with the added `global_` flag.

aligrudi:
Olof-Joachim Frahm <notifications@github.com> wrote:
> I've uploaded [this commit](https://github.com/Ferada/rope/commit/6f7cacc9f261430740086748ae644ccd942ed578) with two failing tests, one just fails to refactor correctly, the other raises an exception because(?) the source code doesn't exactly match the AST, i.e.:
> 
> ````python
>     with open("test") as file1, open("test") as file2:
>         print(file1, file2)
> ````
> 
> fails to parse; there is a second (nested) `with` node in the AST, but in `_Source.consume`/`patchedast.py:643` the `self.offset` is at the end of the line and couldn't match a second, non-existent `with` anyway.
> 
> The other, simpler case is a regular nested `with`, i.e. extracting a method from the following code (the `print` line):
> 
> ````python
>     with open("test") as file1:
>         with open("test") as file2:
>             print(file1, file2)
> ````
> 
> will refactor (and miss the two variables) to:
> 
> ````python
>     def extracted():
>         print(file1, file2)
> 
>     with open("test") as file1:
[...]

PR #398, opened 2021-09-19
Nested with, issue 104

climbus:
https://github.com/python-rope/rope/issues/104

It was tricky issue and need an ugly hack.

I've added support for list of tokens. I think it is more obvious solution compared to adding pseudo object and next if condition: https://github.com/python-rope/rope/blob/master/rope/refactor/patchedast.py#L126

The tricky one was in python2, where nested with and comma separated has identical ast. 

If you have more elegant solution, feel free to send me.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1,0 +1,28 @@
+{
+  "files": [
+    "CONTRIBUTORS.md"
+  ],
+  "imageSize": 100,
+  "contributorsPerLine": 7,
+  "contributorsSortAlphabetically": false,
+  "badgeTemplate": "[![All Contributors](https://img.shields.io/badge/all_contributors-<%= contributors.length %>-orange.svg?style=flat-square)](#contributors)",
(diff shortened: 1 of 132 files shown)


lieryan:
Hi @climbus, thank you for working on this.

If it will significantly simplify this ticket, then I think we should just drop support for Python 2.7.

I've been considering for a while that 0.20.0 might just be the last version of Rope to explicitly support Python 2.7. 

Python 2.7 is already EOL for quite some time and the 1% of people who still uses with Python 2.7 can just download an older version of Rope. AFAICT, 0.20.0 has covered pretty much all of Python 2.7 features, and 2.7 isn't going to grow new features any time soon :sweat_smile:. 

We'll need a Python 2.7 EOL planning to block 2.7 users from accidentally upgrading to incompatible version.

lieryan:
I've opened https://github.com/python-rope/rope/issues/400 to start the ball rolling for dropping support for Python 2.7.

climbus:
> If it will significantly simplify this ticket, then I think we should just drop support for Python 2.7.

By now it isn't a big problem. Few lines of code. 

I'll add my thoughts in the issue you added.

climbus:
@lieryan could you merge this to master? I have to resolve new conflicts.

lieryan:
Hi @climbus thanks for working on this issue. I've merged this PR with a couple additional changes.

Issue #293, opened 2020-03-03
Rename global var affects list comprehension


matmarczak:
With code:
```
some_var = 1
compr = [some_var for some_var in range(10)]
```
When renaming either `some_var` from global or `some_var` from list comprehension scope both variables get renamed.

lieryan:
Copying the investigating text from #249:

> I took a look at this, it seems like there are a number of issues at play here:
>
> 1. `pyscopes.py::_HoldingScopeFinder` currently assumes that a single logical line of code (i.e. a python statement) can only belong only to a single variable scope, but list/generator expression and lambdas are expressions that can create their own variable scope, so this assumption is incorrect. It may take quite a bit of internal rewrite of `_HoldingScopeFinder` and friends to fix this part of the issue.
>
> 2. currently we don't create a variable scope for List/Generator Comprehension, this is in line with Python 2, where Comprehension leaks its loop variables to the surrounding scope, but it's incorrect for Python 3 code
>
> 3. The loop variable of a List/Generator Comprehension currently aren't considered an assignment statement, which means that unless you have an assignment for the same name elsewhere, rope can't resolve the variable (#318)
>
> Probably the quickest way to fix this is to fix that issue 3 first. That should make rope treat comprehension with Python 2 scope semantics (i.e. leaky loop variable), so renames will work most of the time. In Python 3 though, renames will be a bit overzealous if your loop variable shadows a name in the surrounding scope. We can fix the scoping issue later on.

PR #422, opened 2021-09-27
Comprhensions with scope

climbus:
# Description

Added scope for comprehension expressions
Part of https://github.com/python-rope/rope/issues/293

- [ x] I have added tests that prove my fix is effective or that my feature works
- [ x] I have updated CHANGELOG.md


ropetest/pyscopestest.py
```diff
@@ -330,3 +342,56 @@ def test_getting_defined_names_for_modules(self):
...
+
+
+def suite():
+    result = unittest.TestSuite()
+    result.addTests(unittest.makeSuite(PyCoreScopesTest))
+    return result
+
+
+if __name__ == "__main__":
+    unittest.main()
```
lieryan: As we are using `pytest` now, it is no longer necessary to create these Test Suites and main() constructs.

lieryan:
Hi @climbus, thank you, this is really great work.

PR #425, opened 2021-09-29
Refactor to use implicit super()

lieryan:
# Description

Modernise code and use implicit super() syntax.


lieryan:
Doesn't work in Python 2 :(

But good to know that UseFunction refactor works really well for this type of refactoring.

PR #432, opened 2021-10-07
Issue 270

### LINKED ISSUE #270: rope sometimes fails to rename imports with "as"
I've found a minimal, reliable reproducer for a problem renaming with imports that have "as".

First, here's the code I'm using to test:
``` python
rope_test.py:

#!/usr/bin/python3 -tt

import os.path

import rope.base.project
from rope.base import libutils
from rope.refactor.rename import Rename


PROJECT_DIR=os.path.abspath('.')


def main():
    project = rope.base.project.Project(PROJECT_DIR)
    mod = project.find_module('json')
    renamer = Rename(project, mod, None)
    changes = renamer.get_changes('newjson')
    #project.do(changes)
    print(changes.get_description())

if __name__ == '__main__':
    main()
```

Second, here's a test data file that works as expected:
``` python
test.py:

import json as stdlib_json
import json.utils
import json.utils as stdlib_json_utils

print(stdlib_json.dumps({'test': 'case'}))
```

If you drop them into a directory together, and run rope_test.py, you get the expected renames:
``` patch
[pts/3@peru /var/tmp/rope_bug]$ ./rope_test.py                                                            (16:02:31)
Renaming <json> to <newjson>:


--- a/test.py
+++ b/test.py
@@ -1,5 +1,5 @@
-import json as stdlib_json
-import json.utils
-import json.utils as stdlib_json_utils
+import newjson as stdlib_json
+import newjson.utils
+import newjson.utils as stdlib_json_utils
 
 print(stdlib_json.dumps({'test': 'case'}))
```

But, if I comment out all the imports except for the problematic one like this:
``` python
#import json as stdlib_json
#import json.utils
import json.utils as stdlib_json_utils

print(stdlib_json.dumps({'test': 'case'}))
```

Then rope does not rename anything in the file:
```
[pts/3@peru /var/tmp/rope_bug]$ ./rope_test.py                                                            (16:04:46)
Renaming <json> to <newjson>:



[pts/3@peru /var/tmp/rope_bug]$                                                                           (16:04:49)
```

So there's some issue detecting ```import [...] as``` correctly but the presence of other imports to be renamed may inadvertantly mask the bug.

climbus:
# Description

Quick fix #270 

I think it needs some rewriting, but I need some time for investigation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - #398, #104 Fix parsing of nested `with` statement/context manager (@climbus)
 - #391, #376 Fix improper replacement when extracting attribute access expression with `similar=True` (@climbus)
 - #396 Fix improper replacement when extracting index access expression with `similar=True` (@lieryan)
+- #270 Fix rename import statement with dots and as keyword (@climbus)
 
 ## Misc
 
--- a/rope/base/worder.py
+++ b/rope/base/worder.py
@@ -3,6 +3,8 @@ import keyword
 
 import rope.base.simplify
 
+MINIMAL_LEN_FOR_AS = 5
+
 
 def get_name_at(resource, offset):
     source_code = resource.read()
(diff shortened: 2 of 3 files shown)


lieryan:
I'm not quite cozy with how the `MINIMAL_LEN_FOR_AS` is implemented, feels like there must be a better way, but it seems to work fine. Good enough for the purpose, at least, this Worder module also has other weird issues, so we'll probably need to figure out a better way to re-do this module anyway. Maybe a regex-based approach could work better, or maybe look into existing proper lenient Python code parsing libraries.

climbus:
Thanks. I'm not happy with this solution. I planed to think about this one more time. I was wondering if it can be done by ast parser.

PR #444, opened 2021-11-12
Simplify this string, looks like it was a result of the formatter before.

yilei:


lieryan:
Thank you for taking the time to make this change, we have a number of these implicit string concatenation issues after black auto-reformatting.

PR #474, opened 2022-05-20
Rename AutoImport.rope_project back to AutoImport.project

lieryan:
# Description

The change in 89341d4742ac2414051fd671acad8ad48eb9ccf6 broke backward compatibility because it changed what `AutoImport.project` refers to. It was a rope Project but that commit changed it into an autoimport Package.

This PR puts the `.project` back to where it was originally.


lieryan:
FYI, @dryobates

PR #477, opened 2022-05-26
Fix windows

bageljrkhanofemus:
# Description

On windows, the python sys.path includes random files. I do not have windows and 100% of development is on linux but it failed on pylsp's CI. I filtered out files from the search, but cannot easily test this code. I am also unsure of why files are included in sys.path on windows. 

## Fixes 

autoimport on windows.

### DIFF
--- a/rope/base/utils/pycompat.py
+++ b/rope/base/utils/pycompat.py
@@ -34,7 +34,6 @@ except NameError:  # PY3
     def get_ast_with_items(node):
         return node.items
 
-
 else:  # PY2
 
     string_types = (basestring,)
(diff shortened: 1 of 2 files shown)


lieryan:
Instead of checking for not a file, couldn't `filter_folder` just check for directory instead:

    def filter_folders(folder: pathlib.Path) -> bool:
        return (
            folder.is_dir() and
            folder.as_posix() != "/usr/bin"
        )

?

Also, I did find it strange that python-lsp-server has a pytest.exe in its sys.path; adding a folder check in rope might be a hack around this, but it's not necessarily a bad idea either for rope to just ignore garbage/unusable paths here.

PR #480, opened 2022-06-01
Check whether `assignments` attribute exists in `all_star_list`

thomkeh:
# Description

I was getting this error:

```
AttributeError: 'ImportedName' object has no attribute 'assignments'
```
on line 44 in `rope/refactor/importutils/module_imports.py` when I tried to move a function to a different module.

This change fixes that error, but I'm not sure why it happened in the first place.

# Checklist:

### DIFF
--- a/ropetest/refactor/importutilstest.py
+++ b/ropetest/refactor/importutilstest.py
@@ -1873,6 +1873,33 @@ class ImportUtilsTest(unittest.TestCase):
         pymod = self.project.get_pymodule(self.mod)
         self.assertEqual(expected, self.import_tools.organize_imports(pymod))
 
+    def test_organizing_imports_all_star_assigned_name_alias(self):
+        code = expected = dedent("""\
+            from package import name_one, name_two
+
+
+            foo = ['name_one', 'name_two']
+            __all__ = foo
+        """)
+        self.mod.write(code)
+        pymod = self.project.get_pymodule(self.mod)
+        self.assertEqual(expected, self.import_tools.organize_imports(pymod))
+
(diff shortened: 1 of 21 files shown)


ropetest/refactor/importutilstest.py
```diff
@@ -1900,6 +1900,20 @@ def test_organizing_imports_all_star_no_infinite_loop(self):
...
+        code = expected = dedent("""\
+            from mod1 import *
+
+            from mod2 import __all__
+        """)
+        self.mod3.write(code)
+
+        pymod = self.project.get_pymodule(self.mod3)
+        self.assertEqual(expected, self.import_tools.organize_imports(pymod))
+
```
lieryan, line 1916: @thomkeh I added this case, which seems to simulate a possible scenario on how you might end up with the error you are seeing. Can you check if the issue in your codebase is caused by something similar?

thomkeh: Hm, I'm definitely importing some constants, like your
```
some_name = 1
```
and then I add it to `__all__` (and import it with star import somewhere else), but I don't think I have something like `from mod2 import __all__`. And I don't think I'm putting anything in `__all__` that wasn't defined in that file.

thomkeh: In case that was a bit confusing. Here an example:

mod1.py
```
__all__ = ["PI"]
PI = 3.14
```

mod2.py
```
from mod1.py import *
```

lieryan:
@all-contributors add @thomkeh for bug, code

Issue #243, opened 2018-06-12
modify default config.py


dwSun:
Can I have a modified version of default config.py for all my projects? I don't want to copy it every time I start a new project.

soupytwist:
I believe this isn't possible at the moment. If you can contribute this, we would welcome a PR. It could read a simple .roperc file in the user's home dir and override the default_config, for instance.

bageljrkhanofemus:
I'll fix this after #473, you can assign it to me. I'll use a global configuration file.

lieryan:
Hi @bageljrkhanofemus, thanks for looking into this. You're welcome to take this issue, I've assigned this ticket to you.

bagel897:
Implemented in rope 1.5.0

PR #492, opened 2022-06-18
global configuration

bagel897:
# Description

Global config support

Fixes #(issue)
#243 

- [x] I have made corresponding changes to user documentation for new features - needs to specify where the file is per platform

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -4,6 +4,7 @@ Rope supports the following configuration formats
 
 1. pyproject.toml
 2. config.py 
+3. pytool.toml
 
 pyproject.toml 
 --------------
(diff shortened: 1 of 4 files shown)


docs/configuration.rst
```diff
@@ -44,6 +45,10 @@ Additionally, you can run an executable function at startup of rope.
         # Do whatever you like here!
 
 
+pytool.toml 
+-----------
+If neither a config.py or a pyproject.toml is present, rope will use a pytool.toml in the users configuration directory.
```
lieryan, line 50: Where is the "user's configuration directory"? I think we should document the default location, or how the user can get the location here, for example something like:

    If neither a config.py or a pyproject.toml is present, rope will use a pytool.toml in the user's configuration directory (i.e. `${XDG_CONFIG_HOME:-~/.config}/pytool.toml`).

Issue #504, opened 2022-08-25
Autoimport docstring support


bageljrkhanofemus:
Add docstrings to search results. This would be per-name, which causes the following issues:
1. Parsing for both import and ast needs docstring support - import should be trivial and even if ast doesn't work it'll be an improvement.
2. All the databases need a new schema with descriptions
3. The result tuple needs to be updated, which would break things that depend on tuple expansion.

The benefit however, is that we can see docstrings in autoimport results - which I think is a win.

PR #515, opened 2022-11-03
initial docstring support

bagel897:
#504 
## BREAKING CHANGES:
1. Schema change on database (fixed by auto-versioning)
2. SearchResult (only for the full API, not the simple one)

## Other concerns:
- Repeatedly stores module docstrings (we should probably move it to a separate table) 
- Increases the DB size (5mb vs 20mb same env)
- There's some limitations on what we can and cannot get docstrings for (see typing). Since we use AST for these its incomplete and not always accurate (due to decorators or such)
- Slightly increases cache time 1.33 sec to 1.43 sec on large env 
## Preview
image

Side note: I really love the ORM you made, its very cool
- [x] Database migrator (aka check if it has the right columns and wipe otherwise)
- [ ] I have added preferences to enable/disable this (blocked by #516)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 # **Upcoming release**
-
+- #516 Autoimport Now automatically detects project dependencies and can read TOML configuration
 - #733 skip directories with perm error when building autoimport index (@MrBago)
 - #722, #723 Remove site-packages from packages search tree (@tkrabel)
 - #738 Implement os.PathLike on Resource (@lieryan)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # **Upcoming release**
+
 - #516 Autoimport Now automatically detects project dependencies and can read TOML configuration
+- #515 Autoimport now gathers docstrings of imported names and modules (@bagel897).
 - #733 skip directories with perm error when building autoimport index (@MrBago)
 - #722, #723 Remove site-packages from packages search tree (@tkrabel)
 - #738 Implement os.PathLike on Resource (@lieryan)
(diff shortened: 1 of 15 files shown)


lieryan:
Hi bagel, thank you for the contribution.

Since this database is actually really just a "cache"/"index" of the files on disk, I think we don't really need a full blown migration. So, rather than checking whether the database columns matches, we can just have a table to store metadata for the database file, e.g. the version of rope that created the database. If the rope version doesn't match, then just wipe the tables. 

For convenience during development of rope itself, where version number may not reflect actual act lchanges, we can just wipe the database when the hashes the files in rope/contrib/autoimport directory changes, but even a manual wipe would have been fine.

bagel897:
Good point - the tests are in memory so they should work but I'll look into the version table

lieryan:
Hi Bagel, thanks for this, the changes for collecting documentation looks reasonable to me. However, I'm not quite seeing the reason why it is necessary to index the docstrings, maybe you can help me understand here.

Autocompletions need to index names because it needs fast search of the python names as the user types, but once you get the possible matches, wouldn't it be possible to just retrieve the docstring when asked by the editor in real time rather than preemptively returning all docstrings? For example, LSP >= 3.16.0 allows lazy resolution of an autocompleted name's documentation.

Resolving in real-time would allow for more complex resolution strategies as well that wouldn't really be easy during indexing time, for example, if a name is aliased or re-exported by another module, or you can retrieve the documentation for global object instances by inferring the type of the object and finding the docstring of its class.

bagel897:
That is a valid point - this can't get more than like 40% of docstrings so an import based solution would be better for these. On the other hand, I think the increased DB size and time isn't too bad, so it may be worth it to add it here as a baseline and let other projects build atop it (perhaps they could import if its missing/blank?). Also importing has the arbitrary code execution issue while this is AST based.

PR #516, opened 2022-11-03
Autoimport autodetection

bagel897:
# Motivation
Say you have a large environment with a package (IE: Pytorch) that you mark as a dependency. While you plan to use pytorch, its dependents consume a lot of unnessecary space. Additionally, it doesn't make sense to directly import those packages (since it breaks abstraction). 
# Description

Autodetects dependencies
This makes autoimport much much faster (1.33 sec vs 0.28 sec on a large env)
Pyproject.toml support
Will deprecate setting underlined per-function/directly w/o going through prefs (feedback?) 

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -53,15 +53,15 @@ Additionally, you can run an executable function at startup of rope.
 pytool.toml
 -----------
 If neither a config.py or a pyproject.toml is present, rope will use a pytool.toml.
-It follows the exact same syntax of the pyproject.toml.
+It follows the exact same syntax as ``pyproject.toml``.
 
-- Mac OS X: ``~/Library/Application Support/pytool.toml.``
+- Mac OS X: ``~/Library/Application Support/pytool.toml``.
 - Unix: ``~/.config/pytool.toml``` or in $XDG_CONFIG_HOME, if defined
 - Windows: ``C:\Users\<username>\AppData\Local\pytool.toml``
 
 
 Options
-=======
+-------
 .. autopytoolconfigtable:: rope.base.prefs.Prefs
 
 Autoimport Options
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -14,12 +14,11 @@ Will be used if [tool.rope] is configured.
 
     [tool.rope]
     split_imports = true
-    [tool.rope.autoimport]
-    underlined = false
     autoimport.aliases = [
         ['dt', 'datetime'],
         ['mp', 'multiprocessing'],
     ]
+    autoimport.underlined = false
 
 config.py
 ---------
@@ -54,9 +53,9 @@ Additionally, you can run an executable function at startup of rope.
 pytool.toml
 -----------
 If neither a config.py or a pyproject.toml is present, rope will use a pytool.toml.
-It follows the exact same syntax as ``pyproject.toml``.
+It follows the exact same syntax of the pyproject.toml.
 
-- Mac OS X: ``~/Library/Application Support/pytool.toml``.
+- Mac OS X: ``~/Library/Application Support/pytool.toml.``
 - Unix: ``~/.config/pytool.toml``` or in $XDG_CONFIG_HOME, if defined
 - Windows: ``C:\Users\<username>\AppData\Local\pytool.toml``
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -14,6 +14,8 @@ Will be used if [tool.rope] is configured.
 
     [tool.rope]
     split_imports = true
+    [tool.rope.autoimport]
+    underlined = false
 
 config.py
 ---------
@@ -56,8 +58,13 @@ It follows the exact same syntax of the pyproject.toml.
 
 
 Options
--------
+=======
 .. autopytoolconfigtable:: rope.base.prefs.Prefs
+
+Autoimport Options
+------------------
+.. autopytoolconfigtable:: rope.base.prefs.AutoimportPrefs
+
 
 Old Configuration File
 ----------------------
--- a/rope/base/prefs.py
+++ b/rope/base/prefs.py
@@ -15,10 +15,10 @@ from rope.base.resources import Folder
 @dataclass
 class AutoimportPrefs:
     underlined: bool = field(
-        default=False, description="Cache underlined (private) modules"
-    )
+        default=False, description="Cache underlined (private) modules")
     # memory: bool = field(default=None, description="Cache in memory instead of disk")
     # parallel: bool = field(default=True, description="Use multiple processes to parse")
+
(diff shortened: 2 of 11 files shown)


lieryan:
@bagel897 other than Database versioning, is there any other things that this PR will need to resolve before it is ready to merge?

bagel897:
I'm going to be very sporadic on the PR, sorry.
I want to deprecate the underlined setting and make it specified in toml but am unsure of the logistics - we need a way to recreate dbs with new settings and versioning and remove all the existing arguments to functions. It might be too ambitious for little gain, though

lieryan:
No worries, if you are not already working on it, I might pick the task for that database schema versioning so the DB versioning probably should have its own ticket.

bagel897:
I'll move the remaining stuff to future PRs

bagel897:
@lieryan  Sorry for the miscommunication - this PR should be ready.

bagel897:
@lieryan I finally got this one ready, thoughts?

ropetest/contrib/autoimport/deptest.py
```diff
@@ -0,0 +1,54 @@
...
+from rope.contrib.autoimport.sqlite import AutoImport
+
+
+@pytest.fixture
+def project(request, tmp_path: Path) -> Iterable[Project]:
+    doc = request.param
+    if doc is not None:
+        file = tmp_path / "pyproject.toml"
+        file.write_text(doc, encoding="utf-8")
+        print(file, doc)
```
lieryan, line 16: There shouldn't be any printing during test:

```suggestion
```

ropetest/contrib/autoimport/deptest.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import Iterable
+
+import pytest
+
+from rope.base.project import Project
+from rope.contrib.autoimport.sqlite import AutoImport
+
+
+@pytest.fixture
+def project(request, tmp_path: Path) -> Iterable[Project]:
```
lieryan, line 11: There's already a `project` fixture in `ropetest/conftest.py` with a slightly different definition. I'd prefer to avoid having multiple overlapping definitions of fixtures without good reason here.

If we really just want to create pyproject.toml by default, then perhaps make an autouse fixture.

rope/base/prefs.py
```diff
@@ -206,6 +214,8 @@ class Prefs:
             Can only be set in config.py.
         """),
     )
+    autoimport: AutoimportPrefs = field(
+        default_factory=lambda: AutoimportPrefs(), description="Preferences for Autoimport")
```
lieryan, line 218: Is the `lambda` necessary here? Can't it just be: `default_factory=AutoimportPrefs`?

rope/contrib/autoimport/sqlite.py
```diff
@@ -113,25 +133,28 @@ def __init__(
                 autoimport = AutoImport(..., memory=True)
         """
         self.project = project
+        self.prefs = self.project.prefs.autoimport
         project_package = get_package_tuple(project.root.pathlib, project)
         assert project_package is not None
         assert project_package.path is not None
+        if underlined is not None:
+            self.prefs.underlined = underlined
```
lieryan, line 141: Don't we need to copy the `self.prefs` here if we're going to mutate it? Otherwise it would change the setting globally and in other instances of `AutoImport` that shares the same prefs.

lieryan (changes requested):
Thanks @bagel897, excellent as usual. There are a couple notes here that should be looked into.

ropetest/contrib/autoimport/deptest.py
bagel897: fixed in a05f7d53

rope/base/prefs.py
bagel897: fixed in a05f7d53

rope/contrib/autoimport/sqlite.py
bagel897: Good point - I ran a deep copy. Maybe in the future, it should automatically reload on pyproject.toml changes? Would you want me to include that in this PR?

ropetest/contrib/autoimport/deptest.py
bagel897: Hmm - this one has to write the pyproject before the project is created (since that is when the preferences are parsed). If the project fixture used tmp_path, this would be doable, but the directory isn't created until the fixture in this case. Should I rename this fixture to pyproject?

lieryan: We can make the `project` depends on `pyproject_toml` fixture. The default implementation of `pyproject_toml` could just be a no-op (i.e. no `pyproject.toml`), but tests that wants to have one can override the fixture.

bagel897: Done in 0a94b920

rope/base/prefs.py
```diff
@@ -12,6 +12,14 @@
 from rope.base.resources import Folder
 
 
+@dataclass
+class AutoimportPrefs:
+    underlined: bool = field(
+        default=False, description="Cache underlined (private) modules")
+    memory: bool = field(default=None, description="Cache in memory instead of disk")
+    parallel: bool = field(default=True, description="Use multiple processes to parse")
```
lieryan, line 20: Thinking about this further on the config.

First, I don't think `parallel` should be user-configurable. I'd have preferred to not make parallelization configurable at all, but if there's use case single-threaded mode, IMO it's generally the editor/IDE integrators that should choose the best value for their integration, not the user.

Second, I'm not quite sure whether `memory` should be global or autoimport-specific configuration. While this setting currently only affects autoimport, we may want to expand this to other caches. Users that don't want rope to create any cache files likely would want to just turn off all cached files and users that wanted caches mostly wouldn't care about the different cache files that rope generates. 

It's only a small amount of very advanced users or users that are having very specific problems that would've cared about toggling these settings per internal rope component. Unless there's a strong use case for per-component fine tuning here, I would prefer to keep things that can be configured to a minimum to keep maintenance simpler.

docs/configuration.rst
```diff
@@ -14,6 +14,8 @@ Will be used if [tool.rope] is configured.
 
     [tool.rope]
     split_imports = true
+    [tool.rope.autoimport]
+    underlined = false
```
lieryan, line 18: I'd much prefer to keep a flat namespace here for the config keys. The vast majority of users are going to only have a small amount of config, creating multiple config sections just for a handful of configuration seems overkill.

lieryan: IIRC, toml allows [dotted keys](https://toml.io/en/v1.0.0). Maybe we can have our cake and eat it too, by recommending the dotted form in the examples section instead of the separate section:

```toml
[tool.rope]
autoimport.underlined = true
```

PR #450, opened 2021-11-26
Migrate pytest.ini to pyproject.toml

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

# Description

Please include a summary of the change and which issue is fixed. 

Fixes #414 

# Checklist (delete if not relevant):

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated CHANGELOG.md
- [x] I have made corresponding changes to user documentation for new features 
- [x] I have made corresponding changes to library documentation for API changes

PR #519, opened 2022-11-04
dynamic versioning, reformat pyproject.toml, switch pytest.ini to pyproject.toml

### LINKED ISSUE #414: Migrate pytest.ini to pyproject.toml

bagel897:
Pulls versioning from PR + git - necessary for automatic DB upgrades
We need the tags to work first. Also reformatted the pyproject.toml. Please try versioning it yourself first and let me know if you want me to update CI to match.
replaces #450 
closes #414

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,74 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
+  schedule:
+    - cron: '38 14 * * 5'
+
+jobs:
+  analyze:
+    name: Analyze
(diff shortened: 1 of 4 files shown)


bagel897:
How should I leave credit for #450 - I copied it verbatim?

lieryan:
To retain attribution, you can recreate the git commit by setting @gliptak as author. Easiest way to do this, if you're copying the commit more or less verbatim would be using `git cherry-pick`. 

Otherwise, there are some environment variables you can set when creating the commit in command line to override the author fields: https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables, set them to the value that was on the original commit.

I'm not quite sure how dynamic versioning works, I'll need to play with that a bit too see how I like them, though I would have preferred to keep as much of pyproject.toml as possible to be statically defined, to keep things simple for packaging tools. If it needs a dynamic component for the purpose of database re-generation, it should be the version number set in rope/__init__.py that have additional logic to append dirty/unreleased changes information to the version number, e.g. adding a hash there.

bagel897:
Ok so how dynamic versioning works is:
1. If there's a git tag on the current commit, it uses that version - say v1.4.0 would map to 1.4.0
2. Otherwise, you take the previous git tag and do something with the commit number to get a unique dev tag like 1.4.0-dev12856.r204  

The first part means it'd be possible to create a tag and have automatic releases rather than have to bump the version manually. 
The second part helps for developing stuff like auto import without having to manually wipe the DB every commit. (but this isn't nessecary)
If we don't want to use dynamic versioning, then we should remove setuptools-scm from the build dependencies.

As for the attribution, I'll cherry pick it later, thanks

lieryan:
Closing this as this is now replaced with #526.

PR #534, opened 2022-11-26
Fix pylint and mypy complaints re the walk function in rope.base.ast

edreamleo:
[x] Add *.leo to .gitignore so that ongoing changes to .leo files don't pollute rope's code base.
[x] The walk function should not appear to return a value.
[x] Add `-> None` for mypy.
[x] All unit tests pass.

@lieryan Hmm. The diffs for this PR are correct, but there should have been only two commits. Perhaps I should have done a clean or rebase? Any help cleaning this PR would be appreciated.


edreamleo:
Closed because of unwanted commits. See PR #536 instead.

PR #535, opened 2022-11-26
Fix pylint and mypy complaints re the walk function in rope.base.ast

edreamleo:
Second try, after cleaning my master branch.

[x] Add *.leo to .gitignore so that ongoing changes to .leo files don't pollute rope's code base.
[x] The walk function should not appear to return a value.
[x] Add `-> None` for mypy.
[x] All unit tests pass.


edreamleo:
Drat. Still too many commits.

PR #539, opened 2022-11-27
Disambiguate placeholder classes

edreamleo:
This PR lays the foundation for merging pyobjectsdef.py into pyobjects.py (and eliminating the former).

@lieryan stated the the distinction between the two placeholder classes in pyobjects.py and the corresponding classes in pyobjectsdef.py is crucial. This PR shows that that distinction can be made explicit, thereby making it possible to merge pyobjectsdef.py into pyobjects.py.

- [x] pyobjects.py: Add `PyObject.decorators` property.
- [x] pyobjects.py: Rename `PyModule` class to `PlaceholderPyModule`.
- [x] pyobjects.py: Rename `PyPackage' class to `PlaceholderPyPackage`.
- [x] Use the renamed classes in:
    pyobjectsdef.py,
    contrib/generate.py, importutils.actions.py,
    refactor/move.py, refactor/usefunction.py.


lieryan:
Hi @edreamleo, I don't think merging pyobjects and pyobjectsdef is desirable. It is confusing two very different concepts within the codebase. The definition of an object isn't the same as the object itself.

Calling these classes "placeholder" in pyobjects.py is probably a bit misnomer; they aren't really placeholder in the sense that they aren't a real class. It's more like `pyobjects.PyModule` doesn't actually need to implement anything beyond what's implemented in `_PyModule`. But `pyobjects.*` and `pyobjectsdef.*` are very distinct concepts that shouldn't be merged together.

edreamleo:
@lieryan I just took another look at pyobjects.py and pyobjectsdef.py in master. I see that they are not as I had remembered them, so somehow my assumptions in this PR may have been incorrect.

In short, I'm good with closing this PR.

Issue #551, opened 2022-11-28
Fix minor flake8 complaints


edreamleo:
See PR #552.

flake8 reports the following minor problems.  #544 reports more serious problems.

edreamleo:
Closed via PR #552.

PR #552, opened 2022-11-28
PR for #551: fix minor flake8 complaints

edreamleo:
See #551.

- [x] (Off topic, but harmless) Add .pylintrc for close control over pylint.
- [x] Suppress two flake8 warnings in setup.cfg:
    - [x] `B007: Loop control variable 'whatever' not used within the loop body`.
    - [x] `B011: Do not call assert False. Instead raise AssertionError()`
- [x] Add a blank line at the end of setup.py.
- [x] Fix intermixed tabs and spaces in ropetest/contrib/autoimport/utilstest.py.
- [x] Remove extra blank lines in ropetest/contrib/autoimport/utilstest.py.
- [x] Add a comment that leading blank lines are necessary in a docstring in ropetest/type_hinting_test.py.
- [x] Fix the following flake8 warnings:
  - [x] `B009: Do not call getattr with a constant attribute value`.
  - [x] `E722: Do not use bare except, specify exception instead`.
  - [x] `E741: ambiguous variable name 'l'`.
  - [x] `F541: f-string without any placeholders`.
  - [x] `NQA101: "# noqa" has no violations`.


lieryan:
> Fix intermixed tabs and spaces in ropetest/contrib/autoimport/utilstest.py. I don't understand why neither reindent nor black complained

black is disabled on ropetest folder, because black's treatment of multi-line string with textwrap.dedent() is horrendous and the ropetest folder uses a lot of textwrap.dedent() and multiline strings.

If you want to run black on the ropetest folder, you're going to want to use a [specific patch of black](https://github.com/psf/black/pull/1879), and then [adjust black's force-exclude in pyproject.toml](https://github.com/python-rope/rope/blob/master/pyproject.toml#L76). ropetest has been excluded by default so that people don't screw up all the formatting by running normal black on ropetest folder.

edreamleo:
@lieryan Thanks for the heads up about black. I'll adjust my local "blacken" script accordingly.

Heh, I don't remember what checker complained about the intermixed tabs and spaces.  Maybe pylint.

rope/refactor/move.py
```diff
@@ -706,7 +706,9 @@ def rename_in_module(self, new_name, pymodule=None, imports=False, resource=None
 
     def occurs_in_module(self, pymodule=None, resource=None, imports=True):
         finder = self._create_finder(imports)
-        for occurrence in finder.find_occurrences(pymodule=pymodule, resource=resource):
+        for _occurrence in finder.find_occurrences(
+            pymodule=pymodule, resource=resource
+        ):
             return True
         return False
```
lieryan, line 713: Feels like this code could have been simplified, maybe something like:

```suggestion
        occurrences = finder.find_occurrences(pymodule=pymodule, resource=resource)
        return any(True for _ in occurrences)
```

or

```suggestion
        occurrences = finder.find_occurrences(pymodule=pymodule, resource=resource)
        sentinel = object()
        return next(occurrences, sentinel) is sentinel
```

edreamleo: Rev ad9c23ed contains your second suggestion. In the future I'll use the "Commit suggestion" button.

rope/refactor/patchedast.py
```diff
@@ -394,7 +394,7 @@ def _Constant(self, node):
             self._handle(node, ["..."])
             return
 
-        assert False
+        raise AssertionError(f"unknown node.value: {node.value!r}")
```
lieryan, line 397: This isn't a valid transformation. `assert ...` has a very different semantic than `raise AssertionError`.

`assert` is a comment saying that a certain invariant holds true; and `assert ...` is saying that line is meant to be unreachable. `raise ...` implies that there is a possibility of this being a valid code path, which just happens to raise an exception, which isn't the case here. They have very different intent.

Just because it behaves the same in runtime doesn't mean that it has the same semantic. linters are more often than not just very wrong on this kind of things, so I'd suggest we just turn off the linter warnings on assert.

An alternative could be to use `typing.assert_never()`, but the way the `ast` handle Constant nodes, type checkers isn't very useful here.

edreamleo: @lieryan I like your reasoning!  I'll update this PR accordingly:

- Restore the original asserts.
- Suppress `B011: Do not call assert False. Instead raise AssertionError()` in setup.cfg.

rope/base/codeanalyze.py
```diff
@@ -181,7 +181,7 @@ def logical_line_in(self, line_number):
                 indents = count_line_indents(self.lines.get_line(lineno))
 
     def generate_starts(self, start_line=1, end_line=None):
-        for start, end in self.generate_regions(start_line, end_line):
+        for start, _end in self.generate_regions(start_line, end_line):
```
lieryan, line 184: Having a convention where unused variable have to be prefixed with underscore (e.g. `_varname`) isn't going to be enforceable unless we run a linter scanning for this as pre-commit or as a Github Actions; it's too easy to miss that during manual code review. I'd rather we either disable this warning in local linter or add at least a Github Actions to enforce this.

lieryan:
Hi @edreamleo, I think this is about the right size for a PR to be reviewable, any bigger and this will become hard to review the code. Can we freeze this PR at this stage, and split any further fixes should be created as separate PR?

rope/base/codeanalyze.py
edreamleo: flake8 was the source of the complaints. Is it hard to check commits with flake8?

Similarly, it might be good to add a GitHub action that reindent does nothing on all files.

edreamleo:
@lieryan I agree that this PR should not be bigger.

I am willing to close this PR and start a new PR from scratch with the B007 check disabled. A new PR might be best in the long run if you are uncomfortable with adding leading underscores. It's your call.

I'm not sure about my own preferences. I *think* pylint will catch most blunders re B007, but I wouldn't bet my life on it :-)

edreamleo:
@lieryan There is no need for me to create a new PR. It's easy for me to revert specific lines, and the diffs in this PR will confirm that the changes worked as intended.

edreamleo:
@lieryan After further thought, I'd like to delete the added underscores and suppress B007 for now. This will shorten this PR and delay a final decision for (much?) later.

edreamleo:
@lieryan The diffs now show no added underscores in loop vars.  Imo, this PR should now be good to merge.

rope/base/codeanalyze.py
edreamleo: `B007` is is now disabled. We can revisit this question at our leisure.

lieryan (approved):
LGTM, thanks @edreamleo!

rope/base/codeanalyze.py
lieryan: > reindent does nothing on all files, including those in the ropetest directory.

We have Black running on GitHub Actions for formatting checks. I think that should be sufficient, adding more than one formatter would be undesirable, IMO, as it creates the possibility of conflicting advices from the tools.

edreamleo:
@lieryan You're welcome. I am enjoying our collaboration!

rope/base/codeanalyze.py
edreamleo: Good to know.

Issue #553, opened 2022-11-28
Fix potential crasher in merged PR #549


edreamleo:
Somehow unit tests failed to discover a potential crasher in PR #549.

Happily, flake8 did highlight the problem.

Rev bd36aef3 in the ekr-flake8 fixes the problem.  I'll create a new PR asap for just this one fix.

Issue #545, opened 2022-11-27
Fully cover most files


edreamleo:
pytest reports good coverage for most files, but see below!

- [x] ASAP: Create a unit test that would have failed for PR #549.
    This PR contained a potential crasher in rope/refactor/extract.py that unit tests did not discover!
    PR #554 contains the fix.

<details>

<html><body>

Module | statements | missing | excluded | coverage
-- | -- | -- | -- | --
ropetest\refactor\extracttest.py | 1070 | 20 | 0 | 98%
ropetest\refactor\patchedasttest.py | 1018 | 98 | 0 | 90%
ropetest\projecttest.py | 924 | 36 | 0 | 96%
ropetest\pycoretest.py | 860 | 18 | 0 | 98%
ropetest\refactor\importutilstest.py | 830 | 0 | 0 | 100%
ropetest\contrib\codeassisttest.py | 798 | 4 | 0 | 99%
rope\refactor\patchedast.py | 788 | 95 | 0 | 88%
rope\refactor\extract.py | 767 | 36 | 0 | 95%
ropetest\advanced_oi_test.py | 570 | 6 | 0 | 99%
ropetest\refactor\__init__.py | 557 | 1 | 0 | 99%
rope\base\builtins.py | 549 | 40 | 0 | 93%
ropetest\codeanalyzetest.py | 545 | 22 | 0 | 96%
ropetest\refactor\renametest.py | 544 | 8 | 0 | 99%
rope\refactor\move.py | 503 | 24 | 0 | 95%
ropetest\refactor\movetest.py | 498 | 0 | 0 | 100%
rope\base\pyobjectsdef.py | 490 | 23 | 0 | 95%
ropetest\refactor\inlinetest.py | 472 | 3 | 0 | 99%
rope\base\worder.py | 451 | 17 | 0 | 96%
rope\refactor\inline.py | 409 | 28 | 0 | 93%
rope\contrib\codeassist.py | 407 | 59 | 0 | 86%
rope\refactor\importutils\module_imports.py | 404 | 8 | 0 | 98%
ropetest\builtinstest.py | 370 | 0 | 0 | 100%
ropetest\historytest.py | 336 | 6 | 0 | 98%
rope\base\project.py | 310 | 30 | 0 | 90%
rope\base\codeanalyze.py | 287 | 5 | 0 | 98%
ropetest\pyscopestest.py | 282 | 0 | 0 | 100%
rope\contrib\generate.py | 278 | 5 | 0 | 98%
[...]

PR #554, opened 2022-11-28
PR for #553: Fix potential crasher in _getarguments

edreamleo:
See #553. This PR could be called an emergency fix.

flake8 reported the error. Unit tests did not.

Still unresolved: creating a unit test that should have failed :-)  See #545, which suggest better test coverage.


lieryan:
Thanks @edreamleo, I've added a test that covers this function now.

edreamleo:
@lieryan Thanks for adding the unit test!

BTW, I have just added support in Leo for running flake8 automatically when any file changes. So now I'll be using a work flow more similar to yours.

PR #555, opened 2022-11-28
Fix some python2-isms

### LINKED ISSUE #542: Use fstrings
Rope 1.0 could use f-strings, which are far clearer than .format. I have a script that does most of the work.

lieryan:
Fix a number of python2-ism

Fix #542 and others


edreamleo (commented):
:-)

PR #578, opened 2022-12-06
PR: walk does not return a value

edreamleo:
I was *sure* this change had been merged into master, but apparently not.

- mypy complained that walk should not return a value.
- pylint complained that either all return statements should return a value, or no return statement should return a value.


edreamleo:
@lieryan Is there anything I can do to remove the do-nothing previous 5 commits from further PRs?

I've never used or understood squash or rebase.

lieryan:
I suppose the reason why you keep seeing the do-nothing commits is because these commits are in your local master. Assuming that `origin` is the rope's main repository master, you'll want to `git reset origin/master` to set your local master to the main repository's master. You'll also want to make sure that [your fork](https://github.com/edreamleo/ekr-rope)'s master doesn't contain these additional commit, so after you `git reset` your local branch, you'll want to `git push --force ekr-rope master:master`.

For more complex situations, for example if your master do contain something you want to preserve, you're going to want to use `git rebase` to select the commits to preserve, but if you just want to sync up with main repository, then `git reset` is simpler.

edreamleo:
@lieryan Many thanks for your help.

PR #582, opened 2022-12-09
rope.base.ast and rope.base.astutils

lieryan:
Important changes:

- Rename `rope.base.{astutils->nameanalyze}`
- Remove usage of standard library `ast`
- Make `rope.base.ast.parse()` have identical interface to stdlib `ast.parse()`
- Silence unnecessary flake warnings

#### Rename `rope.base.{astutils->nameanalyze}`

`astutils` is a very bad module name for this, many parts of static analysis can be described as "utilities to work with ast", so it's not really descriptive enough as to what `get_name_levels()` and `_NodeNameCollector` really does. The content of this current module isn't actually generic utilities either, they do something very specific, and so calling it "AST utilities" is also very misleading. 

As nothing else is actually left in `astutils`, might as well not have `astutils`.

#### Remove usage of standard library `ast`

rope code should not use standard library `ast` directly, but only through the `rope.base.ast`

#### Make rope.base.ast.parse() have identical interface to stdlib ast.parse()

This function likely can be removed as well, the handling of newlines should really be dealt with at the file access layer, with standard universal newline handling, not here in `ast.parse()`.

#### Silence unnecessary flake warnings

There are warnings from that aren't really helpful on this module.


lieryan:
@edreamleo other than the `from ... import *` matters (which has been silenced), do you think that this would resolve your other concerns with these two modules have been set up?

We can also rename `rope.base.rast` (rast = Rope's AST) if you still have concerns with the name of the module itself; then we'd be able to just `from rope.base import rast` anywhere in the codebase, and it would never be ambiguous with stdlib's `ast`.

edreamleo:
@lieryan Somehow I did not see your question. My apologies for seeming to ignore it.

I don't understand why you say, "rope code should not use standard library ast directly, but only through the rope.base.ast".

PR #583, opened 2022-12-10
PR: Clean imports and fix pylint complaints

edreamleo:
Cherry pick the following commits from PR #580:

- [x] d643366597 Add comments about *difficult* imports in pynames.py 
- [x] a75ed271f9  Clean up/alphabetize imports
- [x] 849891c7d7  Fix three pylint complaints. All tests still pass.
- [x] 89995312ea  Fix/suppress more pylint warnings 

Merging these commits into master will:

- Reduce the size of PR 580.
- Allow PR 580 to be closed without losing any useful work.

PR #584, opened 2022-12-10
PR: Second round: clean imports and qualifiers

edreamleo:
This PR continues PR #583.  This PR suggests a consistent style for imports, thereby regularizing (minimizing) qualified names. All changes should be straightforward and uncontroversial.  

This PR **adds or maintains** (with new comments) fully qualified names for `builtins`, `change`, `fscommands`, and `wildcards` when those names would clash with other names.

This PR **removes** qualifications for modules (like pyobjects and several others) whose names conflict neither with the names other modules (in stdlib or Rope) nor other vars.

**Note**: The qualification of `ast` remains completely unchanged. Some files use `rope.base.ast`; other files access `rope.base.ast` with an unqualified `ast`.


edreamleo:
@lieryan I plan no more commits to this branch.

rope/refactor/functionutils.py
```diff
@@ -1,7 +1,6 @@
-import rope.base.exceptions
-import rope.base.pyobjects
+# import rope.base.exceptions
```
lieryan, line 1: Unused import? 

```suggestion
```

lieryan:
@edreamleo thanks for doing this.

Maybe rather than adding comments about the full import qualification, we should just document this centrally?

I'd recommend a new section in `docs/contributing.rst` describing the rationale for the seemingly "consistently inconsistent" import style. It should say that we defaults to one level import style, and it should document the list of modules that should use full import qualifications. Then can we remove the inline comments. All this can be done in a separate PR.

rope/refactor/functionutils.py
edreamleo: Oops. I'll remove that comment.

edreamleo:
@lieryan I see you have already updated master, so I'll use your suggested code.

edreamleo:
@lieryan I'm good with this PR now.

We could quibble about exactly when to use full qualification, but who would want to do that :-)

lieryan:
Import style consistency is nice, but I'm personally not too fussed if sometimes we forget about it and a couple non compliant imports slipped in.

If necessary, in the future, we can write a pre-commit/GHA script to enforce the import style. So that it'll always be consistent without needing the comments. 

It seems like it can be a useful capability for `findsimilar`/`restructuring` as well to be able to create search patterns that matches imports based on their style. But since we're all ast experts here, seems like it could even be just a simple standalone script.

lieryan:
> We could quibble about exactly when to use full qualification, but who would want to do that :-)

Agreed, the comments are fine as is for now, it's not a blocker for this PR.

PR #585, opened 2022-12-10
PR: Remove python2 metion of raw_input

edreamleo:
@lieryan This PR is experimental. Your comments, please.

`rope.base.builtins` contains an attempted import of `raw_input`, which exists only in Python2.

The PR suggests using `input` instead, including as a dictionary key.

All tests pass, but I don't really understand this code.

### DIFF
--- a/rope/base/builtins.py
+++ b/rope/base/builtins.py
@@ -12,12 +12,6 @@ from rope.base import (
 )
 
 
-try:
-    raw_input
-except NameError:
-    raw_input = input
-
-
 class BuiltinModule(pyobjects.AbstractModule):
     def __init__(self, name, pycore=None, initial={}):
         super().__init__()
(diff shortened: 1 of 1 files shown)


lieryan:
IIUC, `rope.base.builtins` is basically just a catalog of built-in functions and the attributes that those builtins have. Not quite sure off the top of my head where it's used, but likely it is used by inferencing engine how to infer around builtins functions, these inferences are hard coded into rope since we can't actually inspect the "source code" of the builtins as they are part of the Python runtime and aren't Python functions. 

The raw_input there is basically just telling rope that a built-in called raw_input() exists and that it'll return some string value. This is necessary if the user is analysing some Python 2 code. Given that less than 1% of rope users are Python 2 users, I'm not too worried about keeping support for Python 2 here.

We probably need to audit the list of builtins to compare it to the current latest builtins to see if we're missing any or need to deprecate others. I can at least spot a unicode() function there that also are no longer relevant for Python 3.

edreamleo:
@lieryan Thanks your work on the PR.

PR #586, opened 2022-12-10
PR: Simplify annotations in pynames.py

edreamleo:
`pynames.py` drew my attention because of the statement `from __future__ import annotations`.

Initial attempts to simplify the imports failed because of circular references. Then I remembered the pattern I use throughout Leo:

```python
from typing import Any, TYPE_CHECKING  # And any other type constants.

if TYPE_CHECKING:
    # Import any modules used only by the annotations.
    # Define the "real" (extra) annotations.
else:
    # Set all extra annotations to Any.
    # Note: mypy seems to want each annotation to be on a separate line.
```

This pattern is guarranteed to work provided the annotations do not clash with other symbols in the file. I don't recall a single place in Leo's code base where this pattern caused any problems.

### DIFF
--- a/rope/base/pynames.py
+++ b/rope/base/pynames.py
@@ -1,16 +1,17 @@
-# These imports are tricky. It's easy to cause circular imports.
-from __future__ import annotations
-
-import typing
+from typing import Any, Union, TYPE_CHECKING
 
 import rope.base.pyobjects
 from rope.base import exceptions, utils
 
 
-if typing.TYPE_CHECKING:
-    # pyobjectsdef appears only in annotations.
-    from typing import Union
+if TYPE_CHECKING:
     from rope.base import pyobjectsdef
+
+    PyModule = pyobjectsdef.PyModule
+    PyPackage = pyobjectsdef.PyPackage
+else:
+    PyModule = Any
+    PyPackage = Any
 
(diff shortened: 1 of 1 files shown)


lieryan:
What's the benefit of doing this over the current lazy annotations? Seems like it's also less DRY.

lazy annotations is supported by all python versions that are supported by rope, and one way or another possibly after libraries like pydantic resolved its issues with lazy annotations, while the schedule for re-enabling PEP 563 is undetermined, I think it's still likely going to become default at some point in the future.

Due to the nature of what rope does, rope works with a lot of similar thing that are conceptually very different. Modules in changes management, Modules as definitions, and Modules as references are three completely separate concept that rope had to juggle within different contexts. Removing the prefixes from `PyModule` is going to make the code very hard to follow, as it makes it much harder to distinguish between `pyobjects.PyModule` and `pyobjectsdef.PyModule`.

edreamleo:
@lieryan Hah. I didn't know about lazy annotations. I assumed the `from __future__` statement was a holdover from python 2.

I'll close this PR and issue another PR that removes the newly-add comments at the start of `pynames.py`. Those comments are useless/confusing to anyone who *does* understand lazy annotations.

PR #592, opened 2022-12-12
PR: Remove '@testutils.only_for' for versions less than 3.7

edreamleo:
Python version up to and including 3.6 have reached the end of their life.


edreamleo:
Oops. Some tests got deleted.

PR #594, opened 2022-12-13
Refactor patchedast

Alex-CodeLab:
# Description

Some refactoring.
It should not change any behavior. (perhaps a tiny bit faster )

### DIFF
--- a/rope/refactor/patchedast.py
+++ b/rope/refactor/patchedast.py
@@ -9,7 +9,7 @@ from rope.base import ast, codeanalyze, exceptions
 
 COMMA_IN_WITH_PATTERN = re.compile(r"\(.*?\)|(,)")
 
- 
+
 def get_patched_ast(source, sorted_children=False):
     """Adds ``region`` and ``sorted_children`` fields to nodes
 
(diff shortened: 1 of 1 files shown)


lieryan:
Hi @Alex-CodeLab, thank you for taking the time to create this PR. However, I don't think many of these refactoring are desirable.

1. For a pair of brackets/quotes, then it is better for readability that the way the starting and ending token are pushed to the `children` should match for symmetry. 

    ```python
    # good, both starting and ending token uses append
    children = []
    children.append("{")
    ...
    children.append("}")
    
    # good, both starting and ending token are handled by the comprehension
    children = [
        "{", 
        *do_something_with(node.generators),
        "}"
    ]

   # bad, yes, it may save a couple lines, but it lacks symmetry
   children = ["{"]
   ...
   children.append("}")

    ```

2. In general, `children.append()` should be used to push individual unrelated tokens, `children.extend()` should only be used when adding childrens that are one-to-many or for very closely related token:

    This is good:

    ```python
    # good, as `node.generators` is a list
    children.extend(node.generators)
 
    # fine, "async def" functionally acts like one token, even though it is composed of two words
    children.extend(["async", "def"])

    # bad, this is pushing functionally unrelated token, which breaks hierarchy
    def _excepthandler(self, node):
        ...
        children.extend((self.semicolon_or_as_in_except, node.name))
    ```

3. Don't use side-effects with inline-if:

    ```python
    # bad, inline-if shouldn't have side effect
    children.extend(["async", "def"]) if is_async else children.append("def")

    # good, no side effect
    if is_async:
        children.append("def")
    else:
        children.extend(["async", "def"])
    ```

    # better, if you want to do it with one-liner
    children.extend(["async", "def"] if is_async else ["def"])
    ```

Alex-CodeLab:
1. true, it can be somewhat more readable when it is on multiple lines, however doing it this way is also slightly more performant since it avoids the function calls to append. To me it looks a bit weird (inefficient) to create empty lists and then add elements one by one, I guess it is a matter of taste. But I understand your point.

2. Not sure I understand what you mean ;  the result is the same, right? 

3.  yes, makes sense.

lieryan:
The result is the same, but the readability isn't. There's visual and conceptual hierarchy in the syntax, and the code is easier to comprehend when the way the code is written reflects that hierarchy.

The performance difference between append() and extend() for code in patchedast.py is basically negligible, so readability should be the main factor on how you should structure the code here.

Alex-CodeLab:
thanks. 
(sorry I didnt have time to fix it myself)

lieryan:
No worries, thank you for your contribution.

Issue #544, opened 2022-11-27
Fix serious flake8 complaints


edreamleo:
Flake8 reports many *serious* problems. Here is an edited list, omitting the full error messages:

<details>

```console
.\rope\base\builtins.py:16:51: B006 Do not use mutable data structures for argument defaults. 
.\rope\base\builtins.py:95:68: B006 Do not use mutable data structures for argument defaults.
.\rope\base\builtins.py:242:18: B006 Do not use mutable data structures for argument defaults.

.\rope\base\change.py:63:26: B008 Do not perform function calls in argument defaults.
.\rope\base\change.py:75:28: B008 Do not perform function calls in argument defaults.
.\rope\base\change.py:126:28: B008 Do not perform function calls in argument defaults.

.\rope\base\fscommands.py:106:9: B001 Do not use bare `except:`

.\rope\base\history.py:28:39: B008 Do not perform function calls in argument defaults.
.\rope\base\history.py:55:57: B008 Do not perform function calls in argument defaults.
.\rope\base\history.py:80:45: B008 Do not perform function calls in argument defaults.

.\rope\base\libutils.py:72:42: B008 Do not perform function calls in argument defaults.

.\rope\base\project.py:108:39: B008 Do not perform function calls in argument defaults.

.\rope\base\pycore.py:215:39: B008 Do not perform function calls in argument defaults.

.\rope\base\pyobjectsdef.py:114:20: B009 Do not call getattr with a constant attribute value,

.\rope\base\oi\doa.py:191:17: B001 Do not use bare `except:`

.\rope\contrib\findit.py:16:17: B008 Do not perform function calls in argument defaults.
.\rope\contrib\findit.py:49:60: B008 Do not perform function calls in argument defaults.

.\rope\contrib\fixmodnames.py:27:56: B008 Do not perform function calls in argument defaults.

.\rope\contrib\autoimport\pickle.py:99:60: B008 Do not perform function calls in argument defaults.
.\rope\contrib\autoimport\pickle.py:119:53: B008 Do not perform function calls in argument defaults.
.\rope\contrib\autoimport\sqlite.py:156:35: B006 Do not use mutable data structures for argument defaults.
.\rope\contrib\autoimport\sqlite.py:261:50: B008 Do not perform function calls in argument defaults.
.\rope\contrib\autoimport\sqlite.py:291:50: B008 Do not perform function calls in argument defaults.

[...]

PR #595, opened 2022-12-13
fix most of #544 and some minor refactoring.

Alex-CodeLab:
# Description

fix some pep8 issues 

and some minor refactoring ( using contextlib.suppress )

Fixes #(issue)
#544


lieryan:
Hi @Alex-CodeLab, thanks for taking the time to make this PR.

Just a couple things here, I think rather than creating the NullJobSet inside the method, it would make things simpler and better to make a global DEFAULT_TASK_HANDLER in taskhandle.py:

    DEFAULT_TASK_HANDLER = NullJobSet()

and in functions that needed them:

    def call(self, job_set=taskhandle.DEFAULT_TASK_HANDLER):

that'll reduce the line of code that is needed to instantiate the task handler as well as being much more informative when using the view function parameters function in your editor which will direct the caller to see that `job_set` argument takes something from `taskhandle` module instead of the much less useful `job_set=None`.

lieryan:
Thanks for the contribution @Alex-CodeLab!

PR #601, opened 2022-12-15
Fix strange indirections

edreamleo:
rope/base/oi/soi/pynamesdef.py contains:
```python
from rope.base import pynames
from rope.base.pynames import *
```
The purpose of this code is completely mysterious.

I see that this PR is based on the wrong branch. I'll resubmit.


edreamleo:
I am closing this PR: it was based on the wrong branch.

PR #610, opened 2022-12-16
PR: Remove a pyflakes nit and a flake8 suppression

edreamleo:
We all know the pattern:
```python
except SomeException as e:
    ...
```
In this case, however, `e` isn't used. The existing code adds a `# noqa` comment, but there is no way to turn off the pyflakes complaint:
```console
patchedast.py:925:9: local variable 'e' is assigned to but never used
```
Imo, there is no need to fussily add the `as e:  # noqa`. 

@lieryan I'll be happy to add any comment you like just below the changed line.


lieryan:
Looks fine to me, there's no need for extra comment here.

edreamleo:
@lieryan Thanks!

PR #613, opened 2022-12-16
PR: pyflakes nit 4: unused var

edreamleo:
The fourth and last PR in this round of cleanups.

Other pyflakes complaints may disappear later as part of new PRs.

### DIFF
--- a/rope/contrib/findit.py
+++ b/rope/contrib/findit.py
@@ -118,7 +118,7 @@ class Location:
         self.lineno = occurrence.lineno
 
     def __repr__(self):
-        pymod = self.resource.project.get_pymodule(self.resource)
+        self.resource.project.get_pymodule(self.resource)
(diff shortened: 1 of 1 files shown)


rope/contrib/findit.py
```diff
@@ -118,7 +118,7 @@ def __init__(self, occurrence):
         self.lineno = occurrence.lineno
 
     def __repr__(self):
-        pymod = self.resource.project.get_pymodule(self.resource)
+        self.resource.project.get_pymodule(self.resource)
```
lieryan, line 121:
```suggestion
```

This line should just be removed completely, as it doesn't really do anything; I think it was just a leftover from experimenting with different formats for the `__repr__`.

edreamleo (commented):
@lieryan Thanks for the simplification.

I hesitated to remove the call to `get_pymodule` because it wasn't an entirely trivial change

PR #614, opened 2022-12-16
PR: ekr-grand-reorg: disambiguate class names and simplify qualifiers

edreamleo:
The PR eliminates empty distinctions and makes the *crucial*, *real* distinctions explicit.

**Oh joy**: there is no need to merge any files! We need only:
- Remove name clashes.
- Simply imports.

**Newly explicit distinctions (disambiguated names)**

- `pynames.ParameterNameStub` vs. `pyobjectsdef.ParameterName`
- `pyobjects.AssignedNameStub` vs. `pyobjectsdef.AssignedName`
- `pyobjects.PyComprehensionBase` vs `pyobjectsdef.PyComprehension`
   (PyComprehensionBase is not just a Stub class--it has one method)
- `pyobjects.PyFunctionStub` vs. `pyobjectsdef.PyFunction`
- `pyobjects.PyModuleStub` vs. `pyobjectsdef.PyModule`
- `pyobjects.PyPackageStub` vs. `pyobjectsdef.PyPackage`

For consistency, maybe `ParameterNameStub` should move to `pyobjects`.

**Step 1: Merge pynamesdef and pynames**

1. cff's (global searches) revealed that *only* the `ParameterName` name clashed between `pynames` and `pynamesdef`.
  Therefore, I renamed (in pynames) `ParameterName` to `ParameterNameStub`.
  This is an essential distinction, revealed for the first time in this PR.
  Iirc, I also renamed `PyFunction` to `PyFunctionStub` early on.
2. I then ran unit tests until they passed(!).
  I fixed *one failure at a time*, ensuring that I never got carried away with changes.
  I have used this (unintuitive?) tactics many times before, both here and in Leo.

**Step 2: The Grand Compromise: undo the merge!**

Yes, it *is* possible to merge `pynames` and `pynamesdef`. But without odious imports there is no need to do so!

With the new disambiguated names it was straightforward to restore `pynamesdef`!

**Notes**:

- At no time did I ever encounter circular imports.
- Even with tiny changes, I could not have kept the required changes straight in my head without Leo.
  However, the notes above and the diffs should be clear enough.

**Summary**

This PR:
- Disambiguates *all* class names, using uniform qualifiers everywhere.
- Removes the odious `import *` from `pynamesdef`. All imports are now straightforward.

I am proud of this PR. It took *lots* of work to make the diffs as small as they are. The diffs clearly confirm the renaming listed above.

At long last Rope's qualifiers are simple, uniform, truthful, and informative.

### DIFF
--- a/rope/base/arguments.py
+++ b/rope/base/arguments.py
@@ -100,7 +100,8 @@ def _is_method_call(primary, pyfunction):
     pyobject = primary.get_object()
     if (
         isinstance(pyobject.get_type(), rope.base.pyobjects.PyClass)
-        and isinstance(pyfunction, rope.base.pyobjects.PyFunction)
+        # ### and isinstance(pyfunction, rope.base.pyobjects.PyFunction)
+        and isinstance(pyfunction, rope.base.pyobjects.PyFunctionStub)
         and isinstance(pyfunction.parent, rope.base.pyobjects.PyClass)
     ):
         return True
--- a/rope/base/arguments.py
+++ b/rope/base/arguments.py
@@ -100,7 +100,6 @@ def _is_method_call(primary, pyfunction):
     pyobject = primary.get_object()
     if (
         isinstance(pyobject.get_type(), rope.base.pyobjects.PyClass)
-        # ### and isinstance(pyfunction, rope.base.pyobjects.PyFunction)
         and isinstance(pyfunction, rope.base.pyobjects.PyFunctionStub)
         and isinstance(pyfunction.parent, rope.base.pyobjects.PyClass)
     ):
--- a/rope/base/builtins.py
+++ b/rope/base/builtins.py
@@ -7,6 +7,7 @@ from rope.base import (
     arguments,
     ast,
     pynames,
+    pynamesdef,
     pyobjects,
     utils,
 )
@@ -172,7 +173,7 @@ class _CallContext:
         self.args = args
 
     def _get_scope_and_pyname(self, pyname):
-        if pyname is not None and isinstance(pyname, pynames.AssignedName):
+        if pyname is not None and isinstance(pyname, pynamesdef.AssignedName):
             pymodule, lineno = pyname.get_definition_location()
             if pymodule is None:
                 return None, None
--- a/rope/base/evaluate.py
+++ b/rope/base/evaluate.py
@@ -9,6 +9,7 @@ from rope.base import (
     ast,
     nameanalyze,
     exceptions,
+    pynamesdef,
     pyobjects,
     pyobjectsdef,
     worder,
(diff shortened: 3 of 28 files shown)


edreamleo:
@lieryan This PR is more successful than I dared dream. It is a grand compromise--I think it gives us both what we want.  What do you think?

edreamleo:
@lieryan There is one last nit to resolve before this PR is ready for review.

Just as before, the `ImportedModule` class resides in `pynames`, not `pynamesdef`. However the legacy code misleadingly used the `pynamesdef` qualifier. The diffs show the new (different, correct) qualifier for `ImportedModule`.

I don't care where `ImportedModule` resides. Most of the diffs could be made smaller by moving this class to `pynamesdef` and restoring the qualifier.

What would you like to do? I'm good either way.

edreamleo:
@lieryan *Please* merge this PR. Until this happens I will be wasting my time trying to make other contributions to Rope.

lieryan:
Adding a "Stub" suffix here to *some* objects is conveying **totally incorrect and misleading**, it's doing a disservice to misled people reading the code in this way. And it's extremely confusing why is it `pyobjects.PyClass` but `pyobjects.PyFunctionStub`, there's no rhyme nor reason why one should be called Stub but the other isn't.

These aren't stub classes, they are not placeholders, they are actual, real class and there are fundamental conceptual distinction between pynames/pyobjects and pynamesdef/pyobjectsdef. None of these are "just a stub class" the way you have been interpreting them to be.

It's not that I don't want to compromise here, but the reasoning you put on is based on completely misunderstanding what these classes does in the codebase. I would hope that you would try to understand what the existing codebase is doing first before making large sweeping changes.

If you **really** have to name things differently, then what you would want to do is rename all the classes with `pyobjectsdef.DefinedPyClass`, `pyobjectsdef.DefinedPyFunction`, etc. But while that kind of naming convention is accurate, it is also **redundant** because the containing modules' name (i.e. `pyobjectsdef` vs `pyobjects`) already tells you when the code expects an definition or inference objects. The names are **already** non-ambiguous. The naming of these classes are intended to always be interpreted together with the module names, they aren't supposed to be imported individually.

Nobody would complains that pickle.loads, yaml.loads and json.loads have name clashes, because these functions are intended to be imported with the module name, not by themselves. And these libraries wouldn't become better by removing the name clashes by renaming them into: `pickle.pickle_loads()`, `yaml.yaml_loads()`, `json.json_loads()`.

edreamleo:
@lieryan Closing this PR in favor of your PR #618.

edreamleo:
@lieryan 

> Adding a "Stub" suffix here to some objects is conveying totally incorrect and misleading information, it's doing a disservice to misled people reading the code in this way. And it's extremely confusing why is it pyobjects.PyClass but pyobjects.PyFunctionStub, there's no rhyme nor reason why one should be called Stub but the other isn't.

>These aren't stub classes, they are not placeholders, they are actual, real class and there are fundamental conceptual distinction between pynames/pyobjects and pynamesdef/pyobjectsdef. None of these are "just a stub class" the way you have been interpreting them to be.

This is good stuff. Shouldn't the docstrings for these classes say something along these lines? PR #618 doesn't supply docstrings for any of the classes at issue here.

I agree that the 'Stub' *suffix* is misleading. These classes are the base class for other classes. Do you think a 'Base' *prefix* would clarify matters? Or some other prefix?

Issue #619, opened 2022-12-20
Replace rope.base.RopeNodeVisitor with stdlib.ast.NodeVisitor


edreamleo:
`stdlib.ast.NodeVisitor.visit` calls the method called `self.visit_classname`, where classname is the name of the node class, or `generic_visit` if that method doesn’t exist.

`rope.base.ast.RopeNodeVisitor.visit` does the same thing, except that the names of the methods are `self._classname`. Note that `RopeNodeVisitor` already calls `stdlib.ast.NodeVisitor.generic_visit` when a visitor method doesn't exist.

So, if I understand correctly, the task list is:

- [ ] Rename all visitor methods, replacing `_` with `visit_`.
- [ ] Use (stdlib) `ast.NodeVisitor` as the base class of all of Rope's visitors.

lieryan:
That's right, it's a fairly simple change, just somewhat tedious changes. Note that not all functions in RopeNodeVisitor are AST callbacks. There are some helper functions that are just called from other AST callbacks, but would never be called by visit/generic_visit.

edreamleo:
@lieryan Thanks for your helpful comments.

edreamleo:
I am going to close this issue. I have also closed PR #619.

PR #622, opened 2022-12-20
PR for #619: replace rope.base.RopeNodeVisitor with (stdlib) ast.NodeVisitor

edreamleo:
See #619. This PR renames visitors and disambiguates `stdlib.ast` from `rope.base.ast`.

- [x] Add `import ast` as needed.
- [x] Always uses fully-qualified names for `rope.base.ast.parse` and `rope.base.ast.walk`,
    adding `import rope.base` as needed.
- [x] Replace `_` prefix with `visit_` prefix for all visitors.
- [x] All visitor classes are subclasses of `ast.NodeVisitor` instead of `rope.base.ast.RopeNodeVisitor`.
- [x] Add a prominent print statement to `_PatchingASTWalker.__call__`.
  With the `-s` pytest option, this print statement discovered the last few (hard-to-find) problems.


edreamleo:
@lieryan I'm have to pack for my trip :-) I plan no further work no this PR for now.

edreamleo:
@lieryan I have converted this to a draft. There may be a problem with some lowercase node names.

In particular, _PatchingASTWalker._excepthandler and _ScopeVisitor._excepthandler probably should be renamed.

edreamleo:
@lieryan Rev 47f958f fixes a problem with ast.ExceptHandler and ast.excepthandler nodes.

It appears that there are missing unit tests for these two kinds of nodes.

I worried that there might be other problems with lowercase ast node names. Finding such names is more difficult because the names of those visitors can conflict with other methods that start with underscore.

Searching shows that visitors for the following ast nodes appear nowhere in Rope:
boolop, cmpop, operator, pattern, type_ignore, unaryop, withitem.

Visitors for the following ast nodes are now handled correctly:
alias, arg, arguments, comprehension, excepthandler, expr_context, keyword, match_case.

In short, this PR appears ready for review again.

edreamleo:
@lieryan I'm back from vacation and ready to continue our discussion. Do you have any questions or concerns about this PR?

**About wrapping `rope.base.ast`**

Typical wrappers eliminate `if` statements (equivalently, `try/except` blocks) throughout a program. `rope.base.ast` doesn't do so. Instead, `rope.base.ast` obscures the (important!) difference between `stdlib.parse` and `rope.base.ast.parse`.

**Other possible changes**

`ast.py` could be renamed `rast.py`. In that case, the relevant files would import `rope.base.rast` as usual and the code would use the `rast` qualifier instead of `rope.base.ast`.

edreamleo:
@lieryan After further study, I see no reason to remove the `RopeNodeVisitor` class.

`stdlib.ast.NodeVisitor.visit` is designed to be overridden exactly as in `RopeNodeVisitor.visit`.

Re-imagining `rope.base.ast` as something other than a wrapper is a separate issue, with a smaller PR.

PR #632, opened 2023-01-01
PR: Remove `from ast import *` from rope.base.ast

edreamleo:
Another attempt at removing `from ast import *` from `rope.base.ast`.

- [x] Add `import ast` whenever a module uses `stdlib.ast`.
- [x] Remove all `from rope.base import ast` statements.
- [x] Add `from rope.base.ast import RopeNodeVisitor` as needed,
  removing several different qualifiers from references to `RopeNodeVisitor`.
- [x] Replace `ast.parse` with `rope.base.ast.parse`.
- [x] Replace `rope.base.ast.iter_child_nodes` with `ast.iter_child_nodes`.
- [x] Replace `ast.call_for_nodes` with `rope.base.ast.call_for_nodes`.
- [x] Replace `rope.base.ast` qualifiers by `ast` when referring to `stdlib.ast` nodes.
- [x] In numpydocstrings.py, remove `from rope.base.ast import literal_eval`,
   and qualify `literal_eval` with `ast`.
- [x] Remove several unnecessary imports.

**Summary of changes**

- `ast` always means `stdlib.ast`.
- Use full `rope.base.ast` qualifications for the `parse` and `call_for_nodes` functions.
  Alternatively, we could rename `ast.py` to `rast.py` and use the `rast` qualifier.
- Use no qualification for `RopeNodeVisitor`.

In short, all references to `stdlib.ast` and `rope.base.ast` are uniform and explicit.

**Update**

As noted below, three helper PR's will make most of the changes mentioned above.

### DIFF
--- a/rope/contrib/autoimport/parse.py
+++ b/rope/contrib/autoimport/parse.py
@@ -4,13 +4,14 @@ Functions to find importable names.
 Can extract names from source code of a python file, .so object, or builtin module.
 """
 
+import ast
 import inspect
 import logging
 import pathlib
 from importlib import import_module
 from typing import Generator, List
 
-from rope.base import ast
+import rope.base
 
 from .defs import (
     ModuleCompiled,
@@ -45,7 +46,7 @@ def get_names_from_file(
 ) -> Generator[PartialName, None, None]:
     """Get all the names from a given file using ast."""
     try:
-        root_node = ast.parse(module.read_bytes())
+        root_node = rope.base.ast.parse(module.read_bytes())
     except SyntaxError as error:
         print(error)
         return
--- a/rope/base/arguments.py
+++ b/rope/base/arguments.py
@@ -3,7 +3,6 @@ import ast
 import rope.base.evaluate
 
 
-
 class Arguments:
     """A class for evaluating parameters passed to a function
 
--- a/rope/base/arguments.py
+++ b/rope/base/arguments.py
@@ -1,5 +1,7 @@
+import ast
+
 import rope.base.evaluate
-from rope.base import ast
+
 
 
 class Arguments:
--- a/rope/base/ast.py
+++ b/rope/base/ast.py
@@ -1,5 +1,4 @@
 import ast
-from ast import *  # noqa: F401,F403
 
 from rope.base import fscommands
 
--- a/rope/base/builtins.py
+++ b/rope/base/builtins.py
@@ -1,9 +1,10 @@
 """This module tries to support builtin types and functions."""
+import ast
 import inspect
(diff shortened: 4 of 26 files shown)


rope/contrib/autoimport/parse.py
```diff
@@ -45,7 +46,7 @@ def get_names_from_file(
 ) -> Generator[PartialName, None, None]:
     """Get all the names from a given file using ast."""
     try:
-        root_node = ast.parse(module.read_bytes())
+        root_node = rope.base.ast.parse(module.read_bytes())
```
bagel897, line 49: As far as I am aware, this should use the stdlib ast parser

edreamleo: @bagel897 Thanks for you comment! `rope.base.ast.parse` wraps `stdlib.ast.parse`: it regularizes the `source` argument and catches two exceptions. These seem like reasonable things to do.

@lieryan would like to replace the one remaining call to `stdlib.ast.parse` with a call to `rope.base.ast.parse`.

I see two remaining tasks before marking this PR as ready for review:

1. Get PR #633 accepted, so that `rope.base.ast` will define only two names: `rope_parse` (see below) and `RopeNodeVisitor`.

2. Submit a new PR that does the following:

- Rename `rope.base.ast.parse` to `rope_parse`.
- Remove all qualifications for `rope_parse` by adding
  `from rope.base.ast import rope_parse` as needed.
- Replace `ast.RopeNodeVisitor` with `RopeNodeVisitor`,
  adding  `from rope.base.ast import RopeNodeVisitor` as needed.

The new PR will simplify the appearance of the code, remove all ambiguity between the two `parse` methods, and reduce the diffs of *this* PR.

lieryan:
Closed for same reason as https://github.com/python-rope/rope/pull/635#issuecomment-1369709374. This is not the direction we want to take with `rope.base.ast`.

edreamleo:
@lieryan I think we both want to move on. However, I would push back just a bit more.

I created various PRs thinking that smaller PRs would be more acceptable to you. I didn't mean to badger you. I apologize if you felt put upon.

I still think that removing `from ast import *` is a good idea. Most importantly, removing this statement will help mypy. But adding mypy annotations for ast.AST will likely suffice. Otoh, "wrapping" `rope.base.ast` doesn't seem necessary. In particular, the wildcard import does not eliminate "if" statements like typical wrappers do.

Let's put this discussion in perspective. Only three functions are involved, and I don't much care what they are called, nor do I care much what qualifiers are used, so long as the qualifiers are used consistently.

PR #635, opened 2023-01-03
PR: Rename rope.base.ast.parse to rope_parse

edreamleo:
A PR designed to reduce diffs in PR #632.
- [x] Rename `rope.base.ast.parse` to `rope_parse`, adding `from rope.base.ast import rope_parse` as needed.
- [x] Add comment in `rope.refactor.extract._parse_text` that `ast.parse` refers to `stdlib.ast.parse`.
- [x] Add comment about `call_for_nodes` in `rope.refactor.patchedast`.
   If accepted, PR #633 will eliminate `call_for_nodes` entirely.

**Pitch**

- Removes any possibility of confusion between `stdlib.ast.parse` and `rope.base.ast.parse`.
- Removes unnecessary qualifications for `rope_parse`.
- Prepares for another PR that will remove unnecessary qualifications for `RopeNodeVisitor`.
- Will reduce diffs in PR #632.

### DIFF
--- a/rope/base/ast.py
+++ b/rope/base/ast.py
@@ -4,7 +4,7 @@ from ast import *  # noqa: F401,F403
 from rope.base import fscommands
 
 
-def parse(source, filename="<string>", *args, **kwargs):  # type: ignore
+def rope_parse(source, filename="<string>", *args, **kwargs):  # type: ignore
     if isinstance(source, str):
         source = fscommands.unicode_to_file_data(source)
     if b"\r" in source:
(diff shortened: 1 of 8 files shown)


lieryan:
This change just making the code more confusing. The purpose of `rope.base.ast.parse()` being named identically to `ast.parse()` is because there is zero reason for any of rope's code to use the standard library version in rope's codebase. `rope.base.ast` backports changes from future standard library `ast`; and it may also add some code that are **useful** for rope.

You seem to have a rather unhealthy obsession with the `rope.base.ast` module for some reason, can you just move away from `ast` module for now. These PRs trying to make it into what it isn't is counter-productive and aren't really moving the codebase towards the direction that we needed/wanted.

PR #640, opened 2023-01-03
PR: remove unnecessary eval in generate.py

edreamleo:
A nit, discovered while investigating the unit test framework. The eval does not seem like a security problem :-)

The new code does, however, reinforce the docstring.

Hmm. `create_generate` does not appear anywhere in Rope except for the two unit tests that test it. Nor do the docs mention `create_generate`. What's going on?

### DIFF
--- a/rope/contrib/generate.py
+++ b/rope/contrib/generate.py
@@ -18,7 +18,14 @@ def create_generate(kind, project, resource, offset, goal_resource=None):
     'package'.
 
     """
-    generate = eval("Generate" + kind.title())
+    d = {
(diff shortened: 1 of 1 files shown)


lieryan (approved):

lieryan:
It seems to be used by `ropemode` ([src](https://github.com/python-rope/ropemode/blob/912f21a65f99c145822a8d86a6a3f92c9f83cd1f/src/ropemode/refactor.py#L420-L423)), so it's used by ropevim and ropemacs.

It probably should've been documented, or ropemode should've used a documented interface instead.

PR #642, opened 2023-01-04
PR: Add comments & docstrings related to call_for_nodes

edreamleo:
Per the discussion in PR #633.

### DIFF
--- a/ropetest/refactor/patchedasttest.py
+++ b/ropetest/refactor/patchedasttest.py
@@ -1488,7 +1488,6 @@ class _ResultChecker:
         self.test_case.assertEqual((start, end), node.region)
 
     def _find_node(self, text):
-
         """Find the **outer-most last match** in self.ast.
         
         For example, given the goal:
--- a/ropetest/refactor/patchedasttest.py
+++ b/ropetest/refactor/patchedasttest.py
@@ -1488,6 +1488,22 @@ class _ResultChecker:
         self.test_case.assertEqual((start, end), node.region)
 
     def _find_node(self, text):
+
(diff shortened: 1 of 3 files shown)


ropetest/refactor/patchedasttest.py
```diff
@@ -1488,6 +1488,21 @@ def check_region(self, text, start, end):
...
+        
+        and given the text:
+                
+            (a + b) + (c + (d + e))
+            
+        the outermost last match would be:
+            
+            (c + (d + e))
+
+        """
```
lieryan, line 1505: It would be quite a bit confusing to document _ResultChecker using the similarfinder/restructuring syntax, which is completely unrelated to what this class does. Here's a better bit:

```suggestion
        """
        Find the node in `self.ast` whose type is named in `text`.

        :param text: ast node name

        Generally, the test should only have a single matching node, as it make
        the test much harder to understand when there may be multiple matches.

        If `self.ast` contains more than one nodes that matches `text`, then
        the **outer-most last match** takes precedence.

        For example, given that we are looking for `ast.Call` node:

            checker._find_node("Call")

        and given that `self.ast` is the AST for this code:

            func_a(1, func_b(2, 3)) + func_c(4, func_d(5, 6))

        the outer-most last match would be the ast node representing this bit:

            func_c(4, func_d(5, 6))

        Note that the order of traversal is based on the order of ast nodes,
        which usually, but not always, match textual order.
        """
```

Though, I'd really much prefer for `_ResultChecker` to forbid multiple matches instead of documenting what is supposed to be "undefined behavior".

edreamleo:
@lieryan Thanks for improving the docstrings!

PR #646, opened 2023-01-10
Fix renaming kwargs when refactoring from imports

apmorton:
When converting from imports on a large codebase I encountered many files that failed with errors of the form:

```
Syntax error in file <foo.py> line <123>: expression cannot contain assignment, perhaps you meant "=="?
```

This was traced back to erroneously renaming keyword arguments that conflicted with renamed imports, resulting in code that looked something like the following:

Before
```python
from os import path
foo(path=path.join('a', 'b'))
```

After
```python
import os
foo(os.path=os.path.join('a', 'b'))
```


lieryan:
The fix looks reasonable, thanks @apmorton!

@all-contributors add @apmorton for code contribution

PR #648, opened 2023-01-15
Remove __init__ from import statement when using sqlite autoimport

### LINKED ISSUE #645: Remove __init__ from import statement when using sqlite autoimport
**Describe the bug**

**To Reproduce**

1. Code before refactoring:

    ```python
    # mod1/__init__.py
    def foo():
        pass
    ```

2. And you want to autoimport `mod1.foo` from mod2.py using sqlite-based autoimport:

    ```python
    # mod2.py
    foo()
    ```

3. Currently, it seems that sqlite autoimport would suggest this autoimport:

```diff
--- a/mod2.py
+++ b/mod2.py
@@ -1 +1,2 @@
+from mod1.__init__ import foo
 foo()
```

4. What it should have suggested is this refactoring:

```diff
--- a/mod2.py
+++ b/mod2.py
@@ -1 +1,2 @@
+from mod1 import foo
 foo()
```

Note that the legacy pickle autoimport (which is currently still the default) does the right thing and makes the second suggestion.

 - Project Python version: 3.10
 - Rope Python version: 3.10
 - Rope version: 1.6.0

Additional context: sqlite autoimport and pickle autoimport has completely different implementation of autoimport, not just the storage backend.

bagel897:
# Description

Fixes #645


lieryan:
Thanks for fixing this @bagel897

Issue #644, opened 2023-01-08
Improve the docstring for the @saveit decorator


edreamleo:
See PR #652.

I have spent many happy hours over the past two days studying the `@saveit` (aka `@cacheit`) decorator. [This Engineering Notebook page](https://github.com/python-rope/rope/wiki/ENB-4:-The-@saveit-(@cacheit)-decorator) records these investigations.

@lieryan I have become a fan of this decorator. It deserves an excellent docstring. Here is the first draft. Please suggest improvements!
```python
"""
A decorator that supports the `lazy singleton` pattern.

For each decorated function, the decorator creates the singleton value when
the decorated function is first called, *not* when importing a module.

The decorator injects a `private helper ivar` (name) into `self`.
Only the decorator uses this ivar.

:param func: A function/method that instantiates an object.
    
:return: getattr(self, name), the singleton value computed by func.
"""
```
The ENB makes several suggestions:

`@singleton` seems like a better name than either `@saveit` or `@cacheit`.
This decorator involves startup code, not type inference.

`f"_singleton_{func.__name__}.replace('__', '_')` might be a better name for the private helper ivar.

A *separate* `@factory` decorator would be a good idea if `func` returns a factory (Callable).
The docstrings for `@factory` and `@singleton` would be different, as would their annotations.

@lieryan I would appreciate any guidance or comments about this issue or the ENB.

lieryan:
`@saveit` is basically just a version of the same idea as what modern python now has `@functools.cache/cached_property` in the standard library. `@saveit` was devised back when `functools.cache/cached_property` didn't exist.

I think most (all?) instances of `saveit` could just be replaced with `functools.cached_property`. Note that `@cached_property` was added in Python 3.8, so we still have at least a couple months until we can reliably use it in supported versions of Python. My plan was originally to wait it out until Python 3.7 is EOL before deprecating `saveit`, but if you want to do this now, you'll want to backport the functools decorators.

Note that there are also some code of rope that currently manually implements the caching pattern currently done by `saveit`. Most of these could probably be turned into `@cached_property` as well (e.g. [_ResourceMatcher.compiled_patterns](https://github.com/python-rope/rope/blob/d24294fd4f410c72340d98ae8127855549f93192/rope/base/resources.py#L263-L269) and [_ExtractInfo](https://github.com/python-rope/rope/blob/d24294fd4f410c72340d98ae8127855549f93192/rope/refactor/extract.py#L227-L265), there may be others).

[...]

PR #652, opened 2023-01-27
PR2 for #644: Update docstring for saveit decorator

edreamleo:
PR2 for #644: Improve docstring for `@saveit` decorator.

### DIFF
--- a/rope/base/utils/__init__.py
+++ b/rope/base/utils/__init__.py
@@ -1,28 +1,19 @@
 import sys
 import warnings
+from typing import Callable
 
 
-def saveit(func):
+def saveit(func: Callable):
     """
-    A decorator that supports the `lazy singleton` pattern.
+    A decorator calls func *once*, saves function's value, and returns the
+    saved value for the first and all later calls to func.
 
-    For each decorated function, the decorator creates the singleton value when
-    the decorated function is first called, *not* when importing a module.
+    This decorator is part of Rope's startup code. It has no direct role in
+    Rope's type inference code.
 
-    This decorator injects a `private helper ivar` (name) into `self`.
-    Only the decorator uses this ivar.
+    :param func: a Callable.
 
-    :param func: A function/method that instantiates an object.
-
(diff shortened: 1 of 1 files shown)


edreamleo:
@lieryan I've shortened the proposed docstring. I get it: we don't want to write War and Peace for every little function.

And now that I've spent hours on this docstring, it almost seems as if the code is self-explanatory. However, that's me now talking. Imo, the new docstring would help newbies.

Please let me know if you have any concerns about this PR.

lieryan:
Thanks edreamleo, I think the operative word here that should be retained is "cache" or "memoize" should appear somewhere in the docstring.

I think that could replace all this:

```
    saves function's value, and returns the saved value for the first and all later calls to func.
```

Yes, "cache" isn't a word that normal everyday people are expected to understand, but it's a useful word that most programmers should be familiar with, and a standard terminology gets the point across easier and faster rather than trying to figure out what "saves function's value, and returns the saved value for the first and all later calls to func" means.

Docstring should describe **what** the code does and **why**, not **how** the code was implemented; if people are interested in **how** the function works, they should just refer to the code. We may also want to document the naming convention, the cached value is stored in the underscored name of the function.

We would also want to mention that `saveit` is basically just functools.cache/cached_property.

Maybe we should just backport `cached_property` instead. It is currently possible to combine `functools.lru_cache(0)` with `property` to get the same effect as `cached_property/saveit`, and `lru_cache` is already available since Python 3.2. It should look something like this:

```python
from functools import lru_cache

def cached_property(func):
    """ backport of functools.cached_property """
    return property(lru_cache(1)(func))
```

We can revive the pycompat.py file that you removed some time earlier for this function.

edreamleo:
@lieryan Thanks for these comments. I am enjoying this conversation. I'll say more soon.

I've entered a high-energy phase of my Rope work. I'll say more about *that* soon.

I've marked this PR as a draft. Perhaps the final docstring will reference the upcoming Theory of Operation.

PR #655, opened 2023-02-04
PR: Removed unused __init__() methods

edreamleo:
The `_GlobalVisitor` class is an alias of its base class, `_ScopeVisitor`.

Only `dpo.PyModule.__init__` references `_ScopeVisitor`.

The reference is: `self.visitor_class = _GlobalVisitor`

`_GlobalVisitor.__init__` is unnecessary, and is unlikely ever to be necessary.

The only "real" diff should be to `rope/base/pyobjectsdef.py`.

The spurious diffs should disappear once PR #656 is merged into master and this branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # **Upcoming release**
 
 - #650 Install pre-commit hooks on rope repository (@lieryan)
+- #655 Remove unused __init__() methods (@edreamleo, @lieryan)
 
 
 # Release 1.7.0
--- a/rope/base/oi/doa.py
+++ b/rope/base/oi/doa.py
@@ -28,7 +28,7 @@ def _compat_compare_digest(a, b):
     # Computes the bitwise difference of all characters in the two strings
     # before returning whether or not they are equal.
     difference = 0
-    for (a_char, b_char) in zip(a, b):
+    for a_char, b_char in zip(a, b):
         difference |= ord(a_char) ^ ord(b_char)
     return difference == 0
 
--- a/rope/base/oi/runmod.py
+++ b/rope/base/oi/runmod.py
@@ -101,7 +101,7 @@ def __rope_start_everything():
                     returned,
                 )
                 self.sender.send_data(data)
-            except (TypeError):
+            except TypeError:
                 pass
             return self.on_function_call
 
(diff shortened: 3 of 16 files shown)


edreamleo:
@lieryan To be more clear, the question is, if the black-related check passed *this* branch, why did those same checks pass `master`?

edreamleo:
@lieryan  I just got a notification:

Triggered via push February 4, 2023 06:51
@[edreamleo](https://github.com/edreamleo) edreamleo pushed 88e2d68 [ekr-clean-_GlobalVisitor](https://github.com/edreamleo/ekr-fork-rope2/tree/ekr-clean-_GlobalVisitor)

This message is weird, since *you* pushed 88e2d68 to `master`.  What is going on???

edreamleo:
@lieryan There *might* be a simple explanation:

- From gitk it looks like you might not have pushed to master in about 10 days.
- Perhaps the GitHub actions mandated a later version of black in the interval.
  I had to upgrade black myself several days back to make the black tests pass.
- You are now as surprised as I was several days ago.

Just a guess. Anyway, this version of black passes muster:
```console
python.exe -m black --version
python -m black, 23.1.0 (compiled: yes)
Python (CPython) 3.10.5
```

edreamleo:
@lieryan If you like I can create a PR that blackens all files.

I'll reissue this PR once the new PR is merged into master.

lieryan:
Committer, author, and pusher are three separate concept in Git.

I authored and committed https://github.com/python-rope/rope/commit/88e2d68e02072eb9cc225e56e8860ee43a4e9289, but you pushed the branch to your fork when you synced your fork with the mainline repository.

So the notification seems correct.

Github Action seems to use its own upstream version of black. It's not surprising that the formatting changes, since rope pins to an older, patched branch of black that handles dedent()-ed string better. I don't think we pinned the GHA black, so some differences over time would be expected there until black merged the dedent branch.

I'll look into if there's a way to avoid this issues, maybe by pinning GHA black or something.

edreamleo:
@lieryan Thanks for your comments.

I'll continue to use 23.1.0 for now.  I see that `.pre-commit-config.yaml` specifies black 22.12.0.

I'm glad you're on this. GitHub actions remains mysterious to me :-)

edreamleo:
@lieryan Thanks for generalizing this PR to other classes.

PR #657, opened 2023-02-05
PR for #654: Create Theory of Operation document

edreamleo:
See #654.  The PR will remain a draft while writing proceeds.

My [study branch](https://github.com/edreamleo/ekr-fork-rope2/tree/study) is  the base of this PR.

This PR primarily concerns `index.rst` and `theory.rst`. The other diffs contain tracing statements that may be helpful for the Theory of Operation. These traces should never be merged into master.

**Phase 1: Avoid curse of knowledge**

Write as much as I can *now*, without further study. This phase will describe as much as possible from a newbie's point of view.

**Phase 2: Complete the documentation**

Add traces and whatever else is needed.

Issue #654, opened 2023-02-01
Create a Theory of Operation document


edreamleo:
### Description

This task has been the focus of most of my work.

I'll create a first draft when I am able to do so.

Both Rope and Leo use Sphinx, so updating the website itself should be straightforward.

PR #660, opened 2023-02-07
PR2: Theory of Operation

edreamleo:
See #654. Supersedes PR #657. This PR will remain a draft until Phase 3 is complete.

The branch name is misleading&mdash;this PR will contain *only* two `.rst` files.

- [x] Add theory.rst.
- [x] Add reference to theory.rst in index.rst.
- [x] **Phase 1**: Write introduction from a newbie's point of view.
  - Write the intro *before* further study, to avoid the curse of knowledge.
  - Create a list of questions to be answered in Phase 2.
- [ ] **Phase 2**: Study the code in the study branch (not here).
  - There is no need for `tracing_utils.py` here.
  - See #653.
- [ ] **Phase 3**: Complete the first draft.
- [ ] **Phase 4**: Review with @lieryan.


docs/theory.rst
```diff
@@ -0,0 +1,248 @@
...
+Rope's actual startup code is more complex, but for now I think it's safe to ignore these details.
+
+This code falls into `pdb` only for our test:
+
+.. code-block:: python
+
+    def setUp(self):
+        super().setUp()
+        if self.id().endswith('test_simple_type_inferencing'):
+            g.pdb()
```
lieryan, line 221: Rather than embedding the test name like this, usually if I need to drop down to a debugger for a particular test, I'd just drop a hard breakpoint using either the standard `breakpoint()/PYTHONBREAKPOINT` call or `from IPython import embed; embed()` or `from pudb import set_trace; set_trace()`, and then use a test runner to spawn individual pytest, e.g.

    $ pytest path/to/file.py:TestSomething.test_some_test_name

or using partial search path:

    $ pytest -k some_test_name

docs/theory.rst
```diff
@@ -0,0 +1,248 @@
...
+++++++++++++++++++++++++++++++++++++
+
+
+`ObjectInferTest.setUp` executes a simplified version of Rope's startup logic:
+
+.. code-block:: python
+
+    def setUp(self):
+        super().setUp()
+        self.project = testutils.sample_project()
```
lieryan, line 208: sample_project() really just creates an empty folder and setup a Rope project at a temporary directory, that's all it does. setUp() sometimes would seed additional .py files or python packages if the test needs them.

Note also that there's also an alternative version of this fixture written as pytest-style fixture, rather than unittest-style fixture. Once you get what one does, the other should be fairly straightforward to understand.

It's not really representative of Rope's startup sequence. Actual editor integration would often call one of the project validation methods to additionally ensure that any caches are still valid. Not all operations actually **need** the cache though. Many refactoring operations would just do all the necessary inferences when they needed it.

docs/theory.rst
edreamleo: @lieryan Thanks for these tips. I use a similar workflow now.

edreamleo:
@lieryan Rather than create a discussion, I'd like to say here a few words about my approach.

**I'm happy to go slowly**

I think of this PR as a self-guided graduate course. I'm not sure whether it will take one or two semesters :-)

The absence of documentation (including docstrings) forces me to take at least a brief look at *all* of Rope's code. I don't mind. The Theory of Operation should say a few words about how type inference fits into Rope's code base.

I'd rather not ask for help just now. Again, this approach forces me to take a close look at all the relevant code.

When an Aha happens it's important to document *immediately*, before the previous confusion/ignorance becomes completely impossible to recreate. No, the Theory of Operation won't be a history of my study, but it should discuss all the questions that newbies will have.

edreamleo:

PR #661, opened 2023-02-07
PR: Add rope/base/utils/tracing_utils.py

edreamleo:
`tracing_utils.py` is a simplified version of corresponding utils in `leo/core/leoGlobals.py`. I have used such code for several decades. This file:
- Imports only modules in Python stdlib.
- Has no effect on Rope's code base.
- Contains non-trivial tracing code.
- Is fully annotated and passes mypy, flake8 and pylint checks.

**Pitch**

The Theory of Operation document will almost certainly show traces produced by `tracing_utils.py`, so it would be convenient if this file were an official part of Rope. That way the readers of the Theory of Operation would have the tracing code at hand.

Without this PR, traces would be tempted to import `leo.core.leoGlobals`, but doing so runs afoul of Rope's checkin tests.

@lieryan I can live without this PR, but it would be convenient to have for the reasons given above.

### DIFF
--- a/rope/base/utils/tracing_utils.py
+++ b/rope/base/utils/tracing_utils.py
@@ -1,0 +1,133 @@
+"""
+Tracing and debugging functions from Leo.
+
+Duplicated here so that Rope's commit checks won't fail.
+"""
+
+import os
+import pprint
+import sys
+from typing import Any, List
+
+
+def _caller_name(n: int) -> str:
+    """Return the name of the caller n levels back in the call stack."""
+    try:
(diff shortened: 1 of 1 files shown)


edreamleo:
@lieryan I converted this back to a draft because there are problems with only new function, `get_ctor_name`.

We do need a function that aligns traces, but it seems `get_ctor_name` isn't it :-)

edreamleo:
@lieryan I am going to close this PR. I'll reopen it in the unlikely event that traces will be part of the Theory of Operation.

Issue #662, opened 2023-02-11
Fix a valid mypy complaint


edreamleo:
Without a `type:ignore` comment mypy issues the following warning:

     rope\base\pyobjects.py:304: error: "_PyModule" has no attribute "get_name"

The code in question is `pyobjects._PyModule.absolute_name`:
```python
@property
def absolute_name(self) -> str:
    return self.get_name()  # type:ignore
```
This error appears to be valid, but Rope's unit tests do not expose this bug.

Here is the class hierarchy.  All classes appear in pyobjects.py:
```
class _PyModule(PyDefinedObject, AbstractModule):
class PyDefinedObject:
class AbstractModule(PyObject)
class PyObject:
```
None of these classes has a `get_name` member.

Four (passing) unit tests call `pyobjects._PyModule.absolute_name` indirectly:
```python
test_repr_pyobjectsdef_pymodule
test_repr_pyobjectsdef_pymodule_without_associated_resource
test_repr_pyobjectsdef_pypackage
test_repr_pyobjectsdef_pypackage_without_associated_resource
```
These tests will fail if `pyobjects._PyModule.absolute_name` raises an exception.

These tests pass because the tests exercise only the `pyobjectsdef.PyModule`  and `pyobjectsdef.PyPackage` classes. These classes:
- are subclasses of `_pyobjects._PyModule`.
- *do* have `get_name` members.

For example, here is the class hierarchy of `pyobjectsdef.PyModule`:
```
class PyModule(pyobjects.PyModule):
class PyModule(_PyModule):  # pyobjects._PyModule
```
**Summary**

mypy has exposed a potential crasher.

lieryan:
> mypy has exposed a potential crasher.

I don't think there is a practical risk of crashing here because _PyModule is actually an abstract class, it's never meant to be instantiated. Moreover `absolute_name()` is currently only meant to be used for `__repr__`, which is only used for debugging or interacting with Rope objects in REPL. 

If this has been a production code, I'd have made this an abc to enforce that all subclasses implements get_name(), but I wasn't too keen on making it a hard requirement for subclasses to implement a debugging helper method.

edreamleo:
@lieryan I agree, this is no big deal.

PR #663, opened 2023-02-11
PR for #662: fix mypy complaint

edreamleo:
See #662.  This PR fixes *all* mypy complaints involving pyobjects.py:

- In `rope/base/oi/type_hinting/utils.py` replace type comment with annotations.
- Add `AbstractModule.get_name` and `AbstractModule.get_resource`.
- Fully annotate all methods in `class AbstractModule`.

**To do**

- [ ] Verify (@lieryan) that the annotations for `PyObject.get_module` and `AbstractModule.get_module` are correct.

### DIFF
--- a/rope/base/oi/type_hinting/utils.py
+++ b/rope/base/oi/type_hinting/utils.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 import logging
-from typing import Union, TYPE_CHECKING
+from typing import Optional, Union, TYPE_CHECKING
 
 import rope.base.utils as base_utils
 from rope.base import evaluate
@@ -76,7 +76,7 @@ def get_mro(pyclass):
     return class_list
 
 
-def resolve_type(type_name: str, pyobject: PyObj):
+def resolve_type(type_name: str, pyobject: PyObj) -> Optional[PyObject]:
     """
     Find proper type object from its name.
     """
--- a/rope/base/oi/type_hinting/utils.py
+++ b/rope/base/oi/type_hinting/utils.py
@@ -1,10 +1,14 @@
+from __future__ import annotations
 import logging
-from typing import Optional, Union
+from typing import Union, TYPE_CHECKING
 
 import rope.base.utils as base_utils
 from rope.base import evaluate
 from rope.base.exceptions import AttributeNotFoundError
 from rope.base.pyobjects import PyClass, PyDefinedObject, PyFunction, PyObject
+
+if TYPE_CHECKING:
+    PyObj = Union[PyDefinedObject, PyObject]
 
 
 def get_super_func(pyfunc):
@@ -72,8 +76,7 @@ def get_mro(pyclass):
     return class_list
 
 
-def resolve_type(type_name, pyobject):
-    # type: (str, Union[PyDefinedObject, PyObject]) -> Optional[PyDefinedObject, PyObject]
+def resolve_type(type_name: str, pyobject: PyObj):
     """
     Find proper type object from its name.
     """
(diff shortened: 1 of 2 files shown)


edreamleo:
@lieryan Several possible points of discussion:

- An alternative to returning `None` might be to raise `NotImplemented` in various places, especially in `AbstractModule`.
- This PR follows the convention established in the `ekr-mypy` branch:

    `PyObj = Union[PyDefinedObject, PyObject]`

edreamleo:
@lieryan I'd like your guidance for the intended best return types for the two  `get_module` methods this PR changes.

mypy is (for now) happy with having these two modules return almost anything :-)

Presumably mypy will do better checks once more annotations are in place.

edreamleo:

lieryan:
Hi Edward, apologies for not having much replies, at the moment IRL is quite busy, so I hadn't had much chance to look into your pull requests. 

Is there any of these PRs that you are currently looking to target for merging to Rope?

edreamleo:
@lieryan P.S. I am wondering whether Rope has some way of performing "real world" performance tests. The same question appeared in the mypy world. GvR was thrilled that someone could perform performance test on a large codebase.

The question arises because I see no obvious way of finding out whether the extra overhead of the `is_base_x` predicates would be significant. timeit is too localized a measurement.

lieryan:
> super() would still work with two base classes

super() definitely would still work with multiple base classes. Python, like many other languages, uses the [C3 linearization](https://en.m.wikipedia.org/wiki/C3_linearization). 

The super() mechanism is actually designed specifically to allow multiple inheritance, since the old-style way of calling superclass that rope still have in some places (i.e. just calling `ParentClass.__init__(args)`) would often cause problematic behaviour with diamond inheritance (i.e. some superclasses might be called multiple times). super() solves multiple inheritance nicely, making it behave mostly how people expect multiple inheritance should work.

The only issue is that if you're using super() for parent call, you need to do it consistently across the entire the inheritance tree. Rope need to fix those old style superclasses call at some point and just use super() like modern python code. It makes reasoning about the superclasses call much easier. 

> one of those classes is an empty class

This is not a problem with super(). Classes that doesn't implement the method being called using super() just gets skipped in the mro.

lieryan:
> I am wondering whether Rope has some way of performing "real world" performance tests

You can just try opening large open source codebases, for example Django or Twisted. They are some of the largest open source Python codebases I can think off the top of my head at nearly 400kloc and 330kloc of Python respectively (https://codetabs.com/count-loc/count-loc-online.html).

edreamleo:
@lieryan Thanks. I'll consider using Django as a test bed.

In the near term the performance issue is moot. There is an easy way to retain the `isinstance` tests without the `Abstract` classes. Just define three tuples, say `class_types`, `function_types` and `module_types` at the end of `pyobjects.py` whose values are exactly the classes that were formerly subclasses of the three `Abstract` classes. Then test `isinstance(obj, class_types)` instead of `isinstance(obj, AbstractClass)`.

edreamleo:
@lieryan Many thanks for all your helpful comments.

Maybe the better question is whether there are any PRs that *you* would like (eventually) to be merged into Rope :-)

I'll comment later in issue the to-be-created `leo-fork-rope2 issue #2` which will be a ongoing status report and discussion of ideas.

edreamleo:
@lieryan I've just created [leo-fork-rope2 #2](https://github.com/edreamleo/ekr-fork-rope2/issues/2). It summarizes my present plans.

All comments are welcome, but if it's alright with you please create a *new* ["forked" issue](https://github.com/edreamleo/ekr-fork-rope2/issues) for each new topic. That way issue 2 can remain clean.

Btw, I haven't forgotten your remarks about type inference. They may be a crucial part of the puzzle. I'll create a new forked issue on this topic soon.

PR #664, opened 2023-02-12
PR: Simplify Rope's class hierarchy

edreamleo:
This PR collapses the complexity of several type-inference classes:

- Replace all three `Abstract` classes with one new arg to `PyObject.__init__`.
- Use single inheritance for most classes derived from `PyObject` or `PyDefinedObject`.
- `PyDefinedObject` is a subclass of `PyObject`.
  The annotation, `Union[PyObject, PyDefinedObject]`, can be replaced by `PyObject`.

**Note**: The `ekr-mypy-single-inheritance` branch (PR #665) annotates *this* branch, with the usual changes to `rope.base.ast` needed to let mypy do complete checking. Those checks found three missing functions that have now been moved from two `Abstract` classes into the appropriate classes in `pyobjects.py`.

Summary of the changes, in the order they occurred:

- Simplify the `get_base_type` and `get_unknown` functions in `pyobjects.py`.
- Remove unnecessary methods from the three `Abstract` classes.
- Eliminate the `AbstractClass`, `AbstractFunction`, and `AbstractModule` classes.
- Use single inheritance for all important type-inference classes.

**Step 1: Simplify the `get_base_type` and `get_unknown` functions**

- Simplify the `get_base_type` and `get_unknown` functions in `pyobjects.py`.
- Remove a faux helper: the `PyObject._get_base_type` static method.

**Step 2: Remove most methods from the `Abstract` classes**

- Remove most methods from the three `Abstract` classes, retaining 3 methods found nowhere else.
- Move dummy `get_superclasses` from the `BuiltinObject` to its base class, `BuiltinClass`.
  With this change, `BuiltinObject` has the same form as the three `Abstract` classes.

**Step 3: Remove all three Abstract classes**

- Add `rope/base/utils/predicates.py`.
- Replace all tests `isinstance(obj, Abstract*)` with `is_abstract_*(obj)`.
  *Note*: Neither  the `get_base_type` nor the `get_unknown` function requires the `Abstract` classes.
- Remove the three `Abstract` classes entirely.

*Discussion*:

- Rope's inference uses the three `Abstract` classes as markers.
  The legacy pattern is `if isinstance(obj, AbstractX):`
  The new pattern is `if is_abstract_x(obj):`. This pattern is shorter, but is less explicit.
- The `Abstract` classes help init their `PyObject` base classes, but Part 4 shows that this is "faux help".
  The formerly "abstract" classes need only add one new arg when instantiating `PyObject`.

[...]

PR #668, opened 2023-02-14
PR: Remove po._get_base_type. Simplify get_base_type & get_unknown

edreamleo:
- [x] Simplify and annotate the `get_base_type` and `get_unknown` functions in `pyobjects.py`.
- [x] Slightly simplify the docstrings. (See below)
- [x] Remove `PyObject._get_base_type`, a static method. It was a faux helper.
- [x] Use a ternary operator to simplify `PyObject.__init__`.
- [x] Remove the unused `_types` and `_unknown` module-level symbols.

**Discussion**

The `get_base_type` function is an important component of type inference. It deserves to be simplified. The Theory of Operation should discuss the role of the base types.

I don't recall how `get_base_type` came into my view. Maybe I was looking at the functions in `soi.py`. Or maybe I was studying the three `Abstract` classes.

### DIFF
--- a/rope/base/pyobjects.py
+++ b/rope/base/pyobjects.py
@@ -5,9 +5,7 @@ from rope.base import ast, exceptions, utils
 
 class PyObject:
     def __init__(self, type_):
-        if type_ is None:
-            type_ = self
-        self.type = type_
+        self.type = self if type_ is None else type_
 
     def get_attributes(self):
         if self.type is self:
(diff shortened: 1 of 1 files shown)


edreamleo:
@lieryan This PR is ready for review. I considered creating a discussion item for it, but that did not seem necessary. All comments welcome, especially regarding the role of the anchor types in type inference.

The three `AnchorX` classes help init `PyObject`s by calling `super().__init__(get_base_type("X")). PR #664 shows that the `Anchor` classes aren't necessary, but that's for (much?) later. This PR stands on its own.

edreamleo:
@lieryan After writing discussion [#669](https://github.com/python-rope/rope/discussions/669) I dithered briefly about changing this PR.

But no. Let's keep this PR dead simple, and leave other possibilities for later.

rope/base/pyobjects.py
```diff
@@ -63,58 +61,43 @@ def __iter__(self):
...
-    `PyFunction` and `PyClass` for testing if an object is
-    defined somewhere and rope can access its source.  These classes
-    provide more methods.
+    Use:        ``isinstance(pyobject, AbstractFunction)``
+    instead of: ``pyobject.get_type() == get_base_type('Function')``.
 
     """
-    return PyObject._get_base_type(name)
+    global _base_types
+    if _base_types is None:
```
lieryan, line 76: Rather than initializing `base_types` in `get_base_type()`, why not just initialize it in global scope?

You can't do that with `PyObject._types`, because at that point `PyObject` class wasn't defined yet; but if you're moving `PyObject._types` to the global `base_types`, then there wouldn't be that problem, wouldn't it?

edreamleo:
@lieryan Yes, it would be possible to init (underscore) `base_types` in the global scope. I think I considered that.

Your suggestion will work provided the init code follows the `PyObject` class in `pyobjects.py`. I'll change the PR if you like.

The PR's code is kinda like `@utils.saveit`&mdash;the code guarantees that the `PyObject` class has been fully imported regardless of where `get_base_type` appears in the file.

edreamleo:

PR #674, opened 2023-02-25
PR: Fix/suppress all mypy complaints (based on master)

edreamleo:
**Summary**

This PR contains the *minimum* changes (to the `master` branch) needed to fix and suppress all mypy complaints *without* removing the wildcard import in `rope.base.ast`.

This PR:
- Shows that mypy is not adversely affected by the wildcard import in `rope.base.ast`!!
- Suppresses all mypy complaints, mostly with `# type: ignore` comments.
  Comments describe all interesting/unusual suppressions.
- Adds two missing base-class methods whose absence rightly generated mypy errors.

**Details**

- Add `# type: ignore` as needed to suppress warnings about optional imports.
- **rope.base.ast**: mypy rightly complains about an import, but this import does not mypy significantly!
- **rope.base.oi.type_hinting.evaluate**: Suppress complaints about functions decorated with `@multi`.
- **rope.base.oi.type_hinting.utils**: Replace a type comment with a proper annotation.
  The new annotation uses a type alias, `PyObj`, showing that annotations need not be visually daunting.
- **rope.base.prefs**: Suppress all mypy errors by adding `# type: ignore` as the first line of the file.
- **rope.base.project**:  Add  `# type: ignore` to compensate for suppressing mypy in `rope.base.prefs`.
- **Add two missing methods**: `pyobjects.PyObject.get_module` and `pyobjects.PyDefinedObject.get_name`.
  Imo:
    - mypy rightly complained about these methods.
    - Adding these two missing method will help simplify the `Abstract` classes later on.
 - Add (straightforward) annotations of `List` and `Dict` types as needed.
 - **MoveGlobal.get_changes**: Revise existing annotations and suppress complaints.
   Comments suggest actual code revisions, but such changes should wait for a separate PR.

### DIFF
--- a/rope/base/ast.py
+++ b/rope/base/ast.py
@@ -5,6 +5,7 @@ from ast import *  # noqa: F401,F403
 from rope.base import fscommands
 
 try:
+    # mypy right complains: Module "ast" has no attribute "_const_node_type_names"
     from ast import _const_node_type_names  # type:ignore
 except ImportError:
     # backported from stdlib `ast`
--- a/rope/base/ast.py
+++ b/rope/base/ast.py
@@ -5,7 +5,7 @@ from ast import *  # noqa: F401,F403
 from rope.base import fscommands
 
 try:
-    # mypy right complains: Module "ast" has no attribute "_const_node_type_names"
+    # Suppress the mypy complaint: Module "ast" has no attribute "_const_node_type_names"
     from ast import _const_node_type_names  # type:ignore
 except ImportError:
     # backported from stdlib `ast`
--- a/rope/base/ast.py
+++ b/rope/base/ast.py
@@ -5,7 +5,7 @@ from ast import *  # noqa: F401,F403
 from rope.base import fscommands
 
 try:
-    from ast import _const_node_type_names
+    from ast import _const_node_type_names  # type:ignore
 except ImportError:
     # backported from stdlib `ast`
     assert sys.version_info < (3, 8)
--- a/rope/base/pyobjects.py
+++ b/rope/base/pyobjects.py
@@ -19,6 +19,9 @@ class PyObject:
             raise exceptions.AttributeNotFoundError("Attribute %s not found" % name)
         return self.get_attributes()[name]
 
+    def get_module(self):
+        return None
+
     def get_type(self):
         return self.type
 
@@ -224,6 +227,9 @@ class PyDefinedObject:
         while current_object.parent is not None:
             current_object = current_object.parent
         return current_object
+
+    def get_name(self):
+        return None
 
     def get_doc(self) -> Optional[str]:
         if len(self.get_ast().body) > 0:
--- a/rope/refactor/extract.py
+++ b/rope/refactor/extract.py
@@ -1,4 +1,5 @@
 import re
+from typing import Dict
 from contextlib import contextmanager
 from itertools import chain
 
@@ -34,7 +35,7 @@ from rope.refactor import patchedast, similarfinder, sourceutils, suites, usefun
 # There are a few more helper functions and classes used by above
 # classes.
 class _ExtractRefactoring:
-    kind_prefixes = {}
+    kind_prefixes: Dict[str, str] = {}
 
     def __init__(self, project, resource, start_offset, end_offset, variable=False):
         self.project = project
--- a/rope/refactor/move.py
+++ b/rope/refactor/move.py
@@ -7,7 +7,7 @@ based on inputs.
 from __future__ import annotations
 
 import typing
-from typing import Union
+from typing import Optional, List, Union
 
 from rope.base import (
     codeanalyze,
@@ -310,8 +310,8 @@ class MoveGlobal:
 
     def get_changes(
         self,
-        dest: Union[None, str, resources.Resource],
-        resources=None,
+        dest: Optional[Union[str, resources.Resource]],
+        resources: Optional[List[resources.File]] = None,
         task_handle=taskhandle.DEFAULT_TASK_HANDLE,
     ):
         """Return the changes needed for this refactoring
@@ -324,15 +324,21 @@ class MoveGlobal:
           will be applied to all python files.
 
         """
+        # To do (in another PR): Create a new var: dest_resource: resource.Reource
+        # Doing so should remove the type:ignore below.
         if isinstance(dest, str):
             dest = self.project.find_module(dest)
         if resources is None:
             resources = self.project.get_python_files()
+        # The previous guards protect against this mypy complaint:
+        # "Resource" has no attribute "has_child"
         if dest is None or not dest.exists():
             raise exceptions.RefactoringError("Move destination does not exist.")
-        if dest.is_folder() and dest.has_child("__init__.py"):
-            dest = dest.get_child("__init__.py")
-        if dest.is_folder():
+        if dest.is_folder() and dest.has_child("__init__.py"):  # type:ignore
+            dest = dest.get_child("__init__.py")  # type:ignore
+        # The previous guards protect against this mypy complaint:
+        # Item "None" of "Union[str, Resource, None]" has no attribute "is_folder"
+        if dest.is_folder():  # type:ignore
             raise exceptions.RefactoringError(
                 "Move destination for non-modules should not be folders."
             )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - #650 Install pre-commit hooks on rope repository (@lieryan)
 - #655 Remove unused __init__() methods (@edreamleo, @lieryan)
+- #674 Fix/supress all mypy complaints
 
 
 # Release 1.7.0
--- a/rope/base/fscommands.py
+++ b/rope/base/fscommands.py
@@ -11,6 +11,7 @@ import re
 import shutil
 import subprocess
 import typing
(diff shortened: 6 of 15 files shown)


edreamleo:
@lieryan This PR is a milestone in my thinking and workflow:
- It shows that my concerns about the wildcard import were baseless.
- I'm amazed how easy it is to create a PR that merges code from my fork to python-rope.

I plan no further work on this PR. Please let me know if you have any concerns or suggestions.

edreamleo:
@lieryan *Please* comment on this PR in a timely way. I sometimes feel like I'm talking to myself.

If you are too busy to take a close look, just saying *that* would be courteous.

rope/base/pyobjects.py
```diff
@@ -225,6 +228,9 @@ def get_module(self):
             current_object = current_object.parent
         return current_object
 
+    def get_name(self):
+        return None
```
lieryan, line 232: IIUC, this method and get_module() aren't supposed to be called, it's just a dummy implementation to satisfy mypy.

Maybe it should raise NotImplementedError or something like that instead to indicate that only the child class versions of these methods should actually get called.

rope/refactor/extract.py
```diff
@@ -34,7 +35,7 @@
 # There are a few more helper functions and classes used by above
 # classes.
 class _ExtractRefactoring:
-    kind_prefixes = {}
+    kind_prefixes: Dict[str, str] = {}
```
lieryan, line 38: To be more accurate, this could've used Literal to define that it doesn't accept arbitrary string, just a small enumeration of it. Literal is only for Python 3.8 and above though, so we can't use it yet.

rope/refactor/move.py
```diff
@@ -324,15 +324,21 @@ def get_changes(
           will be applied to all python files.
 
         """
+        # To do (in another PR): Create a new var: dest_resource: resource.Reource
```
lieryan, line 327: I wouldn't really recommend creating a new argument here. get_changes() is a public API, pretty much all integrations that supports refactoring would have directly called it. I would be very cautious on changing the API here as it's likely going to break some integrations.

What I'd recommend here is to annotate `is_folder()` to become a [TypeGuard](https://docs.python.org/3/library/typing.html#typing.TypeGuard). Unfortunately, TypeGuard is only available in Python 3.9 and above. 

IMO, we don't need to maintain backwards compatibility with type annotation. If static checks doesn't pass in older Python version, as long as the code is unaffected at runtime, that's perfectly fine to me. 

Alternatively, we could add development dependency on [typing-extensions](https://pypi.org/project/typing-extensions/). which has a backport of `typing.TypeGuard`.

rope/base/ast.py
```diff
@@ -5,7 +5,8 @@
 from rope.base import fscommands
 
 try:
-    from ast import _const_node_type_names
+    # Suppress the mypy complaint: Module "ast" has no attribute "_const_node_type_names"
+    from ast import _const_node_type_names  # type:ignore
```
lieryan, line 9: I'm not a fan of sprinkling so many `type: ignore` everywhere. Is there any way to tell mypy to just ignore untyped modules by default.

edreamleo: @lieryan The `type: ignore` on line 8 of `rope.base.ast` is a special case. I don't know any way to remove it.

There are several ways of telling mypy to ignore untyped modules:
- This PR inserts a `type: ignore` *as the very first line* of `rope.base.prefs`.
- `.mypy` config files supposedly can suppress checks to specific modules.
  However, this capability seems buggy. It seems to suppress initial checks,
  but inferences in other files seem to "bleed into" the supposedly suppressed modules.

One alternative to `type: ignore` is annotating something as `Any`. However, this must substitutes one suppression pattern for another. Let's look at the `type: ignore` statements in this PR:
- Suppress warnings about optional imports. I am not bothered by those suppressions.
  These are the only suppression in `fscommands`, `doa`, `runmod`, `numpydocstrings`, and `project`.
- `oi.type_hinting_evaluate`: suppress (valid) mypy complaints about the `@multi` decorator.
  I haven't looked at the decorated functions in detail, I don't understand why `@multi` is needed,
  so I can't suggest any alternative. My guess: `type: ignore` is likely the only way.
- I know of no easy way to avoid the `type: ignore` in `contrib.autoimport.sqlite`,
  but I didn't try very hard&mdash;this file doesn't interest me at present.
- Afaik, the `type: ignore` suppressions in `refactor.move` can only be removed by "splitting" a name which 
  is assigned values with different types.
  This coding style becomes natural when using mypy. In my experience it always clarifies the code.
  But this PR avoids substantive changes to the code.

**Summary**

Most `type: ignore` suppressions seem harmless. The others seem unavoidable.

I'm not bothered by the `type: ignore` comments. There aren't that many, and I've carefully documented all the unusual suppressions.

rope/base/pyobjects.py
edreamleo: It's true that Rope probably never calls `PyDefinedObject.get_name`.

Imo, however, having to create these two methods indicates a more serious flaw in the Abstract classes.

At present, the draft PR #672 implements only step 1 of a multi-step project. This first step merely moves methods from the `Abstract` classes into what seem to me to be the more natural places, namely the `po.PyFunction` and `po.PyClass' classes (where `po` is my abbreviation for `rope.base.pyobjects`).

Step 2 of PR #672 proposes to make `PyDefinedObject` a subclass of `po.PyObject`. However, this is likely controversial and I have no great attachment to this change.

Oh what the heck. I may as well say here something that I was planning to save for a discussion.

After *weeks* of studying Rope's code I finally see the following:

1. Knowledge is a blessing as well as a curse :-)
   Knowledge helps experienced devs know what's important and what isn't.
2. For newbies the *lack* of knowledge is a curse.
    We can drown in details because we don't know which tidbit might be crucial.
    That's why your comments about inference were so important to me.
    I've just spent several hours revising [forked issue #3](https://github.com/edreamleo/ekr-fork-rope2/issues/3). I could not have done so without your help.
3. Imo, Rope uses a strange form of inheritance&mdash inheritance via mro.
   I understand what's happening now that I have spent days trying to revise it.
   But the class hierarchy will be difficult for newbies to understand without help.
   The Theory of Operation should explain in detail what is going on.
   The alternative is to use a more straightforward form of inheritance.
   The (closed) PR #664 demonstrates that such a simplification straightforward!
   The PR may have gone too far by eliminating the `Abstract` class entirely. It's an open question.

**Summary**

Newbies will struggle to understand the relations between Rope's most important classes.

PR #664 shows that the class hierarchy can be simplified. If this doesn't happen the Theory of Operation should contain a section about how Rope's classes are related.

rope/refactor/extract.py
edreamleo: @lieryan I had forgotten about `Literal`, but adding it seems like small beer.

rope/refactor/move.py
edreamleo: The comment wasn't clear. I wasn't suggesting the api, I was suggesting adding a new *var* (not argument). This change would be an example of what I called "splitting" a var in another reply. mypy will complain about vars that hold values of differing types. It's no big deal whether this ever gets fixed.

edreamleo:
@lieryan Thanks for all your comments. Please resolve the conversations when you are ready. For now I have nothing more to add.

I'll be happy to make any changes you would like. I would appreciate getting this PR merged asap. I would like to use it as the base for future work.

rope/base/pyobjects.py
edreamleo: Meh. The phrase "inheritance by mro" is a fail.

The Aha is that the classes of interest, `po.PyObject`, `po.PyDefinedObject` and the three `Abstract` classes **are (mostly) used as mixin classes**. None of these classes are subclasses of the others, except that the `Abstract` classes are indeed subclasses of `po.PyObject`.

The existing class declarations are:
```python
class BuiltinClass(_BuiltinElement, pyobjects.AbstractClass):
class BuiltinFunction(_BuiltinElement, pyobjects.AbstractFunction):
class BuiltinModule(pyobjects.AbstractModule):
class Generator(pyobjects.AbstractClass):
class Iterator(pyobjects.AbstractClass):
class Lambda(pyobjects.AbstractFunction):
class AbstractClass(PyObject):
class AbstractFunction(PyObject):
class PyObject:
class PyDefinedObject:
class AbstractModule(PyObject):
class PyClass(PyDefinedObject, AbstractClass):
class PyFunction(PyDefinedObject, AbstractFunction):
class _PyModule(PyDefinedObject, AbstractModule):
```
In other words, `PyDefinedObject` *acts* like a `PyObject` even thought `PyDefinedObject` is not a subclass of `PyObject`. Why? Because all subclasses of `PyDefinedObject` are also subclasses of an `Abstract` class so that `PyObject` appears in the mro.

There is a weird kind of logic in this scheme. It's crazy, but imo it's not crazy enough to be right. PR #664 demonstrates that a simpler (more conventional) scheme is feasible.

lieryan:
Thanks for fixing these mypy issues

edreamleo:
@lieryan You're welcome!

edreamleo:
@lieryan I wish to acknowledge that my concerns about `from ast import *` were unfounded. That is, this PR shows that the wildcard import does not impede mypy significantly. It's sometimes a great relief to admit that I was wrong, and this is one of those times :-)

PR #675, opened 2023-02-25
PR: Use single inheritance for Rope's object classes

edreamleo:
PR for [ekr-fork-rope2 #4](https://github.com/edreamleo/ekr-fork-rope2/issues/4)

### DIFF
--- a/rope/base/builtins.py
+++ b/rope/base/builtins.py
@@ -730,6 +730,9 @@ class BuiltinObject(BuiltinClass):
     def __init__(self):
         super().__init__(object, {})
 
+    def get_superclasses(self):
+        return []
+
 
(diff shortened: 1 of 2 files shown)


lieryan:
This change does not look correct to me. 

AbstractFunction classes defines the interface for what it means to be a function, it just provides a default implementation so that subclasses don't have to implement anything if it doesn't need to implement something, ditto with the others. 

Defining empty interface immediately screams that something is wrong with the new version of Abstract classes.

edreamleo:
@lieryan Thanks for your comments! I appreciate your quick reply.

PR #677 supersedes this PR. With your permission I'd like to continue this discussion in PR #677.

PR #678, opened 2023-03-02
PR: Annotate functions/methods related to inference

edreamleo:
This is the start of the my direct study of Rope's inference code.

Notes:

- Only `builtins.Tuple` defines `get_holding_objects`.
  Huh? The calls to this method look like calls to `PyObject.get_holding_objects`.
  It's a pretty strange situation.

### DIFF
--- a/rope/base/pyobjects.py
+++ b/rope/base/pyobjects.py
@@ -1,10 +1,11 @@
+from __future__ import annotations
 from typing import Optional
 
 from rope.base import ast, exceptions, utils
 
 
 class PyObject:
-    def __init__(self, type_):
+    def __init__(self, type_: PyObject):
         if type_ is None:
             type_ = self
         self.type = type_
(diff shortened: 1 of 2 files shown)


rope/base/pyobjects.py
```diff
@@ -22,7 +23,7 @@ def get_attribute(self, name):
     def get_module(self):
         return None
 
-    def get_type(self):
+    def get_type(self) -> PyObject:
```
lieryan, line 26: This is probably going to require a bit of re-thinking and re-architecturing here with `get_type()`. `get_type()` is supposed to return the inferred type of an object, so the intended case here is that it should return some instance of AbstractClass (e.g. int = BuiltinClass, or user-defined class = PyClass). 

However, in the current implementation, when rope cannot infer the type of an object, the current implementation would just return just some empty PyObject as a placeholder, which doesn't look correct/ideal to me. Maybe we could have some sort of UnknownClass instead to more explicitly describe what's happening; and then this method then can instead be annotated as such:

```
def get_type(self) -> AbstractClass:
    ...
```

PR #680, opened 2023-03-05
PR: Remove a do-nothing statement in soi._handle_first_parameter

edreamleo:
The `then-clause` of an `if` statement was `pass`.


lieryan:
Hi @edreamleo, thanks for fixing this.

I wonder what was the intent for this empty block was.

If I had to guess, this was probably supposed to be an if--elif chain and the empty block was supposed to be an early return from the method. 

Not quite sure whether the lack of early return would have caused any issues, but if it's been a decade without an issue, removing it probably would be fine.

PR #688, opened 2023-04-18
Fix autoimport not scanning packages recursively

### LINKED ISSUE #687: Missing symbols in autoimport.db
### Discussed in https://github.com/python-rope/rope/discussions/685

<div type='discussions-op-text'>

<sup>Originally posted by **rstcruzo** April  6, 2023</sup>
Hey everyone. I have been using rope with pylsp for a couple of days and noticed something odd. Autoimport suggestions contain my virtual env packages if I add the virtual env site-packages folder to the python_path preference. That's okay, but seems like rope does not scan all subfolders to generate autoimports.

For instance, I'm using `prompt_toolkit`, and the `prompt_toolkit.prompt` is returned by rope to auto import it, but the `prompt_toolkit.completion.Completer` is not. `prompt_toolkit.history.FileHistory` is listed too. So the behavior is inconsistent.

Seems to me that rope is not able to autoimport from folders inside the prompt_toolkit, since `completion` is a folder, `history` is a python file. Is this expected? or is there any configuration I can set to fix this?

I also checked the autoimport.db file and looks like the folders of prompt_toolkit are missing from there too.</div>

lieryan:
# Description

Change glob pattern in get_files() to scan site-packages recursively.

Fixes #687 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - #650 Install pre-commit hooks on rope repository (@lieryan)
 - #655 Remove unused __init__() methods (@edreamleo, @lieryan)
 - #674 Fix/supress all mypy complaints
+- #687 Fix autoimport not scanning packages recursively (@lieryan)
 
 
 # Release 1.7.0
--- a/rope/contrib/autoimport/utils.py
+++ b/rope/contrib/autoimport/utils.py
@@ -118,7 +118,7 @@ def get_files(
         yield ModuleFile(package.path, package.path.stem, underlined, False)
     else:
         assert package.path
-        for file in package.path.glob("*.py"):
+        for file in package.path.glob("**/*.py"):
             if file.name == "__init__.py":
                 yield ModuleFile(
                     file,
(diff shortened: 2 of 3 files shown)


lieryan:
cc: @bagel897 FYI, this is an issue raised by rstcruzo in #685. I hope I didn't missing anything obvious with this fix here.

bagel897:
Shouldn't it be **.py? That way it'll glob root directory files.

lieryan:
Yeah, that's what I initially thought as well, but apparently, the `glob.glob()` and the `pathlib` glob doesn't really work the same.

Doing `**.py` is an error with pathlib's glob:

```python
In [5]: list(Path(".").glob("**.py"))
ValueError: Invalid pattern: '**' can only be an entire path component
```

`**/*.py` looks unintuitive, but it does actually include the root directory files:

```
In [12]: list(Path(".").glob("**/*.py"))
Out[12]: 
[PosixPath('setup.py'),
...
]
```

OTOH, for `glob.glob()` stdlib module, `**.py` is basically identical to `*.py`:

```python
In [10]: glob.glob("**.py", recursive=True)
Out[10]: ['setup.py']
```

and `**/*.py` works the same as pathlib's glob.

bagel897:
Weird

PR #696, opened 2023-06-10
Add prefer_global_from_import

nicoolas25:
# Description

Adding a `prefer_global_from_import` configuration option for pick the "global" form of import over the module one.

```diff
- from pkg import mod1
- mod1.AClass()
+ from pkg.mod1 import AClass
+ AClass()
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # **Upcoming release**
 
 - #624 Implement `nonlocal` keyword (@lieryan)
+- #696 Introduce the `prefer_global_from_imports` configuration (@nicoolas25)
 
 # Release 1.8.0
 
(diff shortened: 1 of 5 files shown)


lieryan:
Hi @nicoolas25, thank you for the contribution.

Adding this configuration would imply that there will be two conflicting configuration options with the `prefer_module_from_imports` and `prefer_global_from_imports` and both can't be set to True at the same time.

I think the config options should've been merged. 

The `prefer_module_from_imports` config option should be deprecated and the new config option, e.g. `preferred_import_style` should have string values like `auto`, `normal-import`, `from-module`, `from-global`, etc. This new config option should take precedence over the legacy value.

lieryan:
Thanks for contributing this change, @nicoolas25. I've adapted your changes as part of #788.

@all-contributors please add @nicoolas25 for code contribution

PR #714, opened 2023-10-23
Allow sqlite to be accessed from different thread

### LINKED ISSUE #713: Calling Autoimport.connection.execute from differnt thread causes error
**Describe the bug**
When calling any autoimport API that uses `Autoimport.connection.execute()` from a thread different than the one from which the sqlite db was created, we get an error

```
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id <x> and this is thread id <y>.
```

**To Reproduce**

1. Setup

```python

project = Project('.')
autoimport = AutoImport(project, memory=False)
autoimport.generate_cache()  # Generates a cache of the local modules, from the project you're working on
autoimport.generate_modules_cache()  # Generates a cache of external modules

def search_and_print(search_string):
    print(list(autoimport.search_full(search_string)))
```

2. This works
```python
search_and_print('numpy')
```

3. This throws an error
```python
import threading
thread = threading.Thread(target=search_and_print, args=("numpy",))
thread.start()
thread.join()
```

**Why is this relevant?**
rope is used by `python-lsp-server`, which uses `threading`

**Additional context**
The fix is easy. I'll open up a PR

tkrabel-db:
# Description

Remove checking if sqlite db is accessed from a different thread.

Fixes #713 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - #710, #561 Implement `except*` syntax (@lieryan)
 - #711 allow building documentation without having rope module installed (@kloczek)
+- #714 allow autoimport db connection to be used by mutliple threads (@tkrabel)
 
 # Release 1.10.0
 
--- a/rope/contrib/autoimport/sqlite.py
+++ b/rope/contrib/autoimport/sqlite.py
@@ -87,7 +87,7 @@ class AutoImport:
         project: Project,
         observe: bool = True,
         underlined: bool = False,
-        threadsafe: bool = True,
+        check_same_thread: bool = True,
         memory: bool = _deprecated_default,
     ):
         """Construct an AutoImport object.
@@ -100,7 +100,7 @@ class AutoImport:
             If true, listen for project changes and update the cache.
         underlined : bool
             If `underlined` is `True`, underlined names are cached, too.
-        threadsafe : bool
+        check_same_thread : bool
             If true, sqlite3 connection can only be used in the thread it was created in.
         memory:
             If true, don't persist to disk
@@ -129,7 +129,7 @@ class AutoImport:
         self.connection = self.create_database_connection(
             project=project,
             memory=memory,
-            threadsafe=threadsafe,
+            check_same_thread=check_same_thread,
         )
         self._setup_db()
         if observe:
@@ -144,7 +144,7 @@ class AutoImport:
         *,
         project: Optional[Project] = None,
         memory: bool = False,
-        threadsafe: bool = True,
+        check_same_thread: bool = True,
     ) -> sqlite3.Connection:
         """
         Create an sqlite3 connection.
@@ -153,7 +153,7 @@ class AutoImport:
             The project to use for project imports.
         memory : bool
             If true, don't persist to disk.
-        threadsafe : bool
+        check_same_thread : bool
             If true, sqlite3 connection can only be used in the thread it was created in.
         """
         if not memory and project is None:
@@ -163,7 +163,7 @@ class AutoImport:
             db_path = ":memory:"
         else:
             db_path = str(Path(project.ropefolder.real_path) / "autoimport.db")
-        return sqlite3.connect(db_path, check_same_thread=threadsafe)
+        return sqlite3.connect(db_path, check_same_thread=check_same_thread)
 
     def _setup_db(self):
         models.Metadata.create_table(self.connection)
--- a/rope/contrib/autoimport/sqlite.py
+++ b/rope/contrib/autoimport/sqlite.py
@@ -156,7 +156,7 @@ class AutoImport:
             db_path = ":memory:"
         else:
             db_path = str(Path(project.ropefolder.real_path) / "autoimport.db")
-        return sqlite3.connect(db_path)
+        return sqlite3.connect(db_path, check_same_thread=False)
 
     def _setup_db(self):
         models.Metadata.create_table(self.connection)
--- a/rope/contrib/autoimport/sqlite.py
+++ b/rope/contrib/autoimport/sqlite.py
@@ -87,6 +87,7 @@ class AutoImport:
         project: Project,
         observe: bool = True,
         underlined: bool = False,
+        threadsafe: bool = True,
         memory: bool = _deprecated_default,
     ):
         """Construct an AutoImport object.
@@ -94,11 +95,13 @@ class AutoImport:
         Parameters
         ___________
         project : rope.base.project.Project
-            the project to use for project imports
+            The project to use for project imports
         observe : bool
-            if true, listen for project changes and update the cache.
+            If true, listen for project changes and update the cache.
         underlined : bool
             If `underlined` is `True`, underlined names are cached, too.
+        threadsafe : bool
+            If true, sqlite3 connection can only be used in the thread it was created in.
         memory:
             If true, don't persist to disk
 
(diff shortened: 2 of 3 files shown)


lieryan:
Hi @tkrabel-db, thanks for making this PR. I'm not quite sure that exposing this attribute to the editor-side API would be the best way to fix this issue. The editor-side of the integration (i.e. python-lsp-server in this instance) ideally shouldn't have to care about the internal concurrency requirement of rope. Whether or not an API calls the database or not is not part of the rope API, it'll complicate the API if only some calls can be made safely from multiple threads and especially with nothing documenting what is or is not safe.

I did a little bit of reading on this, from my understanding, with the appropriate compile-time options, sqlite3 actually does support multithreading and `check_same_thread=False` just bypasses a cpython-level check? According to https://discuss.python.org/t/is-sqlite3-threadsafety-the-same-thing-as-sqlite3-threadsafe-from-the-c-library/11463, Python 3.11 seems to expose this compile time option now using the DBAPI 2.0 attribute `sqlite3.threadsafety` but it should have already been safe to use sqlite3 from multiple threads in earlier versions of python if you do the compile options detection yourself.

For the highest compatibility, we would need to implement our own serialization mechanism to proxy all calls into a single worker thread. This means every API would need to queue requests to a single worker thread. This is doable, but unless a lot of modern OSes/IDEs that have compiled sqlite3 for Python with multithreading disabled, it seems like it'll be unnecessarily complicated solution. 

If we drop support for configuration where sqlite3 is not compiled with multithreading support, then instead of exposing `check_same_thread` to python-lsp-server, rope could instead set a value for `check_same_thread` after checking the compile time options. I think this would simplify supporting multithreading. This means that if an editor wants to use rope with a Python without sqlite3 multithreading support, it's their responsibility to wrap autoimport in a single worker thread.

tkrabel:
@lieryan thanks for the investigation and the detailed explanation. I am short on time myself right now, but I wanted to add some data points that may help.
`sqlite3.threadsafety` on python 3.8 returns `1`, which means that the [module can be shared across threads, but not connections](https://docs.python.org/3/library/sqlite3.html#sqlite3.threadsafety). IIUC, `AutoImport` is the main entry point for any CRUD operation, and it holds one connection to the data base. So if a user wanted to use rope with `threadsafety = 1`, she'd need to have one `AutoImport` object per thread, right?

lieryan:
With threadsafety = 1, yes, they'll need to have one AutoImport per thread. But another thing to be careful about is when using in-memory database, IIUC, whenever you create an unnamed in memory database in sqlite, that creates a new, empty database. So one AutoImport per thread will mean that additional memory usage and additional scanning time for each thread. That would be undesirable.

If we want to recommend one AutoImport per thread as the best way to multithreading, we'll likely want to change AutoImport to use an explicitly named in-memory database so that they'll all connect to the same database. Creating [named in memory database](https://www.sqlite.org/inmemorydb.html) seems to be quite simple using a filename like so "file:memdb1?mode=memory&cache=shared", if we name the database with something that uniquely identify the rope Project, maybe the hash of `id(project)` or the path to the project root, I believe that should work.

I think I would have preferred one AutoImport per thread better than making use of threadsafety = 2 as that seems like a much simpler threading model to use and understand.

tkrabel:
That makes sense to me. So let's close this PR. I'll address the respective change of namespacing the in-memory DB in a separate PR.

tkrabel:
I was checking multi-threading behaviour on python 3.11, where `sqlite3.threadsafety = 3`, essentially telling that threads can share the module, connectors, and cursors. However, when calling a db API from a different thread, we still get the error explained in https://github.com/python-rope/rope/issues/713. I think `check_same_thread=False` needs to be set to avoid that error, independent of what `sqlite3.threadsafety` says. That said, I interpret `sqlite3.threadsafety` to say if the OSes python version can guarantee threadsafety, while `check_same_thread` is an extra safeguard by `sqlite3` to prevent users from shooting themselves in the foot.

From my POV, it should be up to the user of a sqlite db to determine if they run into thread safety issues or not, but at least they should be able to turn checks off if they deem their use case safe.

We could add a warning to `check_same_thread`, saying that users should know what they do if they turn it off. WDYT @lieryan ?

lieryan:
Can we confirm that we're talking about the same thing here? If by "users" you meant end users, I don't think end users should have to care about the implementation details of rope. If instead by users you meant authors of editors/lsp servers, then I can see the merit for exposing the flag, though I still don't think it's the best way to solve the issue.

If we require each thread to create one AutoImport, all we need to do to handle concurrency is just to use standard sqlite transactions; if we reuse the same connections in multiple threads, we will also need to implement a way to control access to the connection. Calling AutoImport from multiple threads will open up a whole new class of failure scenarios. 

While doing individual database operations should work fine with thread safe sqlite, if there are operations that depends on the atomicity of a sequence of reads and writes and which will require using transaction, reusing the same connection for multiple threads doesn't really make sense anyway. 

If rope is the end application and have full control over how the connection is used from different threads, this might be a-ok, but as rope is a library that can be called from many different clients we don't control, trying to get every implementor to handle concurrency correctly is going to be tricky.

On the other hand, the concurrency issues that can happen if we require one AutoImport per thread rule is mostly going to be the same as the concurrency issues caused by having multiple instances of rope running. For example if you have multiple editors/IDE, or if you run multiple instances of pylsp for the same project. This is a use case we necessarily have to support anyway, so it doesn't create any new failure scenarios.

Also, the sqlite authors stated that while sqlite is threadsafe with the appropriate compile options, they also stated that ["Threads are evil. Avoid them."](https://www.sqlite.org/faq.html#q5) and recommends avoiding threading with sqlite. I would be inclined to heed to the warning.

lieryan:
Ok, I think I have a better idea now. Rather than one AutoImport per thread, I think we can make it so that AutoImport will keep the sqlite connection in a [threadlocal](https://docs.python.org/3/library/threading.html#thread-local-data). If the editor application always calls rope from a single thread, this has negligible impact for them, everything will continue to work as they currently do. But if they call AutoImport methods from multiple threads, AutoImport will initialise an sqlite connection for each thread. In either case, the editor still uses a single instance of AutoImport so there's minimal change needed on the editor side code, but each threads will internally be served by their own sqlite connection, which avoids needing multi threading access to the same connection.

tkrabel:
> Can we confirm that we're talking about the same thing here? 
By "users" I mean everyone that uses that library, but I think this mostly likely will be application developers that want to expose rope's features to an end user, so you can say I mean application / editor developers.

> Also, the sqlite authors stated that while sqlite is threadsafe with the appropriate compile options, they also stated that ["Threads are evil. Avoid them."](https://www.sqlite.org/faq.html#q5) and recommends avoiding threading with sqlite. I would be inclined to heed to the warning.

Noted.

tkrabel:
> Rather than one AutoImport per thread, I think we can make it so that AutoImport will keep the sqlite connection in a [threadlocal](https://docs.python.org/3/library/threading.html#thread-local-data)

I like that approach! It's clean, i.e. the APIs won't change, and we handle the "complexity" of ensuring one connection per thread in the background.

I opened a PR https://github.com/python-rope/rope/pull/720

lieryan:
@tkrabel IIUC, with the one connection per thread, changing `check_same_thread` should no longer be needed, correct? I'm closing this PR under that assumption, let me know if otherwise.

tkrabel-db:
@lieryan that is correct. Thanks for cleaning this up! :)

PR #719, opened 2023-10-28
Use in-memory database with shared cache

tkrabel:
# Description

As discussed in https://github.com/python-rope/rope/pull/714#issuecomment-1783791890, this PR sets the foundation for using `AutoImport` in a multithreading environment by using an in-memory database with shared cache per `Project` if `memory=True`. On-disk databases are not affected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - #710, #561 Implement `except*` syntax (@lieryan)
 - #711 allow building documentation without having rope module installed (@kloczek)
-- #719 Allow in-memory databases being shared across threads (@tkrabel)
+- #719 Allows the in-memory db to be shared across threads (@tkrabel)
 
 # Release 1.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - #710, #561 Implement `except*` syntax (@lieryan)
 - #711 allow building documentation without having rope module installed (@kloczek)
-- #719 Create in-memory databases with shared cache per project to support a one-thread-one-autoimport model in multithreading environments (@tkrabel)
+- #719 Allow in-memory databases being shared across threads (@tkrabel)
 
 # Release 1.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - #710, #561 Implement `except*` syntax (@lieryan)
 - #711 allow building documentation without having rope module installed (@kloczek)
+- #719 Create in-memory databases with shared cache per project to support a one-thread-one-autoimport model in multithreading environments (@tkrabel)
 
 # Release 1.10.0
 
--- a/rope/contrib/autoimport/sqlite.py
+++ b/rope/contrib/autoimport/sqlite.py
@@ -154,10 +154,16 @@ class AutoImport:
         if memory or project is None or project.ropefolder is None:
             # Allows the in-memory db to be shared across threads
             # See https://www.sqlite.org/inmemorydb.html
-            project_hash = hash(project and project.ropefolder and project.ropefolder.real_path)
-            return sqlite3.connect(f"file:memdb{project_hash}:?mode=memory&cache=shared", uri=True)
+            project_hash = hash(
+                project and project.ropefolder and project.ropefolder.real_path
+            )
+            return sqlite3.connect(
+                f"file:memdb{project_hash}:?mode=memory&cache=shared", uri=True
+            )
         else:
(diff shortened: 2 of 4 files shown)


rope/contrib/autoimport/sqlite.py
```diff
@@ -153,7 +153,10 @@ def create_database_connection(
             raise Exception("if memory=False, project must be provided")
         db_path: str
         if memory or project is None or project.ropefolder is None:
-            db_path = ":memory:"
+            # Allows the in-memory db to be shared across threads
+            # See https://www.sqlite.org/inmemorydb.html
+            project_hash = hash(project and project.ropefolder and project.ropefolder.real_path)
+            db_path = f"file:memdb{project_hash}:?mode=memory&cache=shared"
```
tkrabel, line 159: This creates a database file, i.e. writes everything to disk, even though I followed https://www.sqlite.org/inmemorydb.html!

tkrabel:
@lieryan this is ready for review.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - #710, #561 Implement `except*` syntax (@lieryan)
 - #711 allow building documentation without having rope module installed (@kloczek)
+- #719 Allow in-memory databases being shared across threads (@tkrabel)
```
lieryan, line 5:
```suggestion
- #719 Allow in-memory databases being shared across threads (@tkrabel)
```

rope/contrib/autoimport/sqlite.py
```diff
@@ -151,12 +151,13 @@ def create_database_connection(
         """
         if not memory and project is None:
             raise Exception("if memory=False, project must be provided")
-        db_path: str
         if memory or project is None or project.ropefolder is None:
-            db_path = ":memory:"
+            # Allows the in-memory db to be shared across threads
+            # See https://www.sqlite.org/inmemorydb.html
+            project_hash = hash(project and project.ropefolder and project.ropefolder.real_path)
```
lieryan, line 157: Can we avoid using `hash` function here, instead use a more standard hashes in hashlib. While we don't really need the strong security property of hashing, the `hash()` function is tied to the implementation of dict/set rather than being a general purpose hash function.

tkrabel: Please note that I am essentially hashing the string that is returned by `project.ropefolder.real_path`. I do the checks as a fail-safe, but it should not happen that project is not None, but project.ropefolder is ([source](https://github.com/python-rope/rope/blob/ff50dde5dd63a68866145d7844e3219e39cba1d3/rope/base/project.py#L197)). 

I can make the distinctions clearer if you want with another if condition:

```suggestion
            project_hash: int
            if project is None or project.ropefolder is None:
                project_hash = hash(None)
            else:
                project_hash = hash(project.ropefolder.real_path)
```

I added a unit test that checks for the most common use case I think: if we have two different projects, then we get two different in-memory databases, while same projects share the database. 

I find using `hashlib` overkill here, as it doesn't add much value since we pass in `None` or `str` to it.

lieryan: IIRC, if project.ropefolder is None, that means the `.ropeproject` is disabled for that project. In that case, I think we should hash the project's path instead of the ropefolder's path. 

Also, this

```
project_hash = hash(None)
```

does not look right. It meant that when no project is provided, everything will connect to the same in-memory database. The expectation here is that when no project is provided, that it should always create a new, empty database, so the right way to do this might be to just generate a random hash.

tkrabel: I also checked the code and it seems project = None is not allowed when creating an AutoImport instance, and  it is allowed by create_database_connection, but not even used in tests by us. My feeling is to let everything set on fire if somebody uses create_database_connection without specifying a project, but I don't want to be a bad person so I just create a random hash that has the same format as the regular hash :)

rope/contrib/autoimport/sqlite.py
```diff
@@ -151,12 +152,23 @@ def create_database_connection(
         """
         if not memory and project is None:
             raise Exception("if memory=False, project must be provided")
-        db_path: str
         if memory or project is None or project.ropefolder is None:
-            db_path = ":memory:"
+            # Allows the in-memory db to be shared across threads
+            # See https://www.sqlite.org/inmemorydb.html
+            project_hash: int
```
tkrabel, line 158: I thought about moving this into its own function, but that would add another hop for understanding what's happening, increasing obscurity of the code base. So I decided against it. I follow the principle of "small interfaces" and "deep modules"

lieryan:
@tkrabel thank you for creating this PR

PR #720, opened 2023-10-29
Create one database connection per thread

### LINKED ISSUE #713: Calling Autoimport.connection.execute from differnt thread causes error

tkrabel:
# Description

Addresses https://github.com/python-rope/rope/pull/714#issuecomment-1784018856, using a threadlocal to create data base connections.

Fixes #713 

### DIFF
--- a/rope/contrib/autoimport/sqlite.py
+++ b/rope/contrib/autoimport/sqlite.py
@@ -68,7 +68,6 @@ def filter_packages(
 
 
 _deprecated_default: bool = object()  # type: ignore
-thread_local = local()
 
 
 class AutoImport:
@@ -127,6 +126,7 @@ class AutoImport:
                 "`AutoImport(memory=True)` explicitly.",
                 DeprecationWarning,
             )
+        self.thread_local = local()
         self.connection = self.create_database_connection(
             project=project,
             memory=memory,
@@ -169,16 +169,16 @@ class AutoImport:
 
         This makes sure AutoImport can be shared across threads.
         """
-        if not hasattr(thread_local, "connection"):
-            thread_local.connection = self.create_database_connection(
+        if not hasattr(self.thread_local, "connection"):
+            self.thread_local.connection = self.create_database_connection(
                 project=self.project,
                 memory=self.memory,
             )
-        return thread_local.connection
+        return self.thread_local.connection
 
     @connection.setter
     def connection(self, value: sqlite3.Connection):
-        thread_local.connection = value
+        self.thread_local.connection = value
 
     def _setup_db(self):
         models.Metadata.create_table(self.connection)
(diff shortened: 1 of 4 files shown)


rope/contrib/autoimport/sqlite.py
```diff
@@ -67,6 +68,7 @@ def filter_package(package: Package) -> bool:
 
 
 _deprecated_default: bool = object()  # type: ignore
+thread_local = local()
```
lieryan, line 71: IIUC, creating the thread_local at the module-level like this would mean that all instances of autoimports would share the same database and database connection, even though they belong to different instances of AutoImport and more importantly that they belong to different Projects:

```python
project_a = Project("/path/to/project/a")
project_b = Project("/path/to/project/b")
ai_a = AutoImport(project_a, ...)
ai_b = AutoImport(project_b, ...)

assert ai_a.connection is not ai_b.connection
```

That does not look correct to me. I think this can be avoided by moving `thread_local` to be an instance variable initialized in `AutoImport.__init__()`.

lieryan (changes requested):
Hi @tkrabel, thanks for making this PR, there's one issue with the code currently as they are. Once they are fixed, I think it should be ready to go.

rope/contrib/autoimport/sqlite.py
tkrabel: Thanks for pointing that our! You're right. I moved the thread local to the `__init__` and wrote a test that verifies the following passes:

```python
# Test connections

ai1 = AutoImport(Project(".."))
ai2 = AutoImport(Project("."))
ai3 = AutoImport(Project("."))
assert ai1.connection is not ai2.connection
assert ai1.connection is not ai3.connection
```

lieryan:
Looks good to me, thanks for the contribution @tkrabel.

https://github.com/all-contributors add @tkrabel for code contribution

lieryan (approved):

lieryan:
Looks good to me, thanks for the contribution @tkrabel.

@all-contributors add @tkrabel for code contribution

PR #721, opened 2023-10-30
fix #715

### LINKED ISSUE #715: AutoImport's `get_modules` is case insensitive
**Describe the bug**

AutoImport's `get_modules` is case insensitive. Not sure if exactly a bug, but I don't think this behavior fits considering the method is used for full names and not for autocompletion purposes.

**To Reproduce**
```
from pathlib import Path

autoimport = AutoImport(Project(Path()), memory=False)
autoimport.generate_cache()
autoimport.generate_modules_cache()
print(autoimport.get_modules("C"))
```

The code above prints `['calendar']` but `from calendar import C` is not a valid import. In fact, the correct symbol is `c` (lower-case).

The same happens with the `pickle` version of `AutoImport`.

**Editor information:**
 - Project Python version: 3.10.13
 - Rope Python version: 3.10.13
 - Rope version: 1.10.0
 - Text editor/IDE and version: N/A

bagel897:
#715

### DIFF
--- a/rope/contrib/autoimport/models.py
+++ b/rope/contrib/autoimport/models.py
@@ -111,6 +111,7 @@ class Name(Model):
 
     search_by_name_like = objects.where("name LIKE (?)")
 
+    search_by_name = objects.where("name IS (?)")
     delete_by_module_name = objects.where("module = ?").delete_from()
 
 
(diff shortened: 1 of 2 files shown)


lieryan:
Thank you for looking into this @bagel897.

PR #723, opened 2023-11-01
Remove site-packages from packages search tree

### LINKED ISSUE #722: AutoImport.generate_modules_cache can be speeded up by 2x
**Describe the bug**

I played around with `AutoImport.generate_modules_cache`, as it is [perceived to be slow](https://github.com/python-lsp/python-lsp-server/issues/431#issuecomment-1783792855). I logged what packages get imported from a conda env I created and linked to the project, and I seems there are many unnecessary duplicate entries added to the database.

You can see the imports in this file: [sorted_packages_with_site_packages.txt](https://github.com/python-rope/rope/files/13229466/sorted_packages_with_site_packages.txt).

I see the following pattern of duplicate entries in that file.
```
Name(name=<import_name>, modname=<mod_name>, package=<package>, ...)
Name(name=<import_name>, modname="site-packages."<mod_name>, package='site-packages', ...)
```
One example of a duplicate:
```
Name(name='BaseName', modname='jedi.api.classes', package='jedi', source=<Source.SITE_PACKAGE: 4>, name_type=<NameType.Class: 7>)
Name(name='BaseName', modname='site-packages.jedi.api.classes', package='site-packages', source=<Source.SITE_PACKAGE: 4>, name_type=<NameType.Class: 7>)
```

From that, it seems the issue is that we don't exclude the top level `site-packages` directory itself from our search tree, which treats it as its own package and hence every package inside of it is double counted.

**To Reproduce**

1. Change code in `sqlite.py`

```diff
diff --git a/rope/contrib/autoimport/sqlite.py b/rope/contrib/autoimport/sqlite.py
index eb7c27de..42447edc 100644
--- a/rope/contrib/autoimport/sqlite.py
+++ b/rope/contrib/autoimport/sqlite.py
@@ -371,6 +371,7 @@ class AutoImport:
             return
         self._add_packages(packages)
         job_set = task_handle.create_jobset("Generating autoimport cache", 0)
+        end_names = []
         if single_thread:
             for package in packages:
                 for module in get_files(package, underlined):
@@ -383,9 +384,11 @@ class AutoImport:
                 get_future_names(packages, underlined, job_set)
             ):
                 self._add_names(future_name.result())
+                end_names.append(future_name.result())
                 job_set.finished_job()
 
         self.connection.commit()
+        return end_names
 
     def _get_packages_from_modules(self, modules: List[str]) -> Iterator[Package]:
         for modname in modules:
```

2. Run the code from an env that has the changes from (1) applied

```python

project = Project(".")
autoimport = AutoImport(project, memory=True)
autoimport.generate_modules_cache()
```

3. Look at the result

tkrabel:
# Description

Remove `site-packages` from the python package search tree to avoid duplicate entries in the autoimport database. This also speeds up generating the modules cache by a factor of ~2.

Fixes #722 

### DIFF
--- a/rope/contrib/autoimport/sqlite.py
+++ b/rope/contrib/autoimport/sqlite.py
@@ -529,6 +529,7 @@ class AutoImport:
             chain(*self._execute(models.Package.objects.select_star()).fetchall())
         )
         existing.append(self.project_package.name)
+        existing.append("site-packages")
         return existing
 
     def _removed(self, resource):
--- a/rope/contrib/autoimport/sqlite.py
+++ b/rope/contrib/autoimport/sqlite.py
@@ -529,7 +529,6 @@ class AutoImport:
             chain(*self._execute(models.Package.objects.select_star()).fetchall())
         )
         existing.append(self.project_package.name)
-        existing.append("site-packages")
         return existing
 
     def _removed(self, resource):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - #710, #561 Implement `except*` syntax (@lieryan)
 - #711 allow building documentation without having rope module installed (@kloczek)
+- #722 speed up module cache generation by factor of 2 (@tkrabel)
 
 # Release 1.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # **Upcoming release**
 
 - #710, #561 Implement `except*` syntax (@lieryan)
-- #711 allow building documentation without having rope module installed (@kloczek)
-- #722 speed up module cache generation by factor of 2 (@tkrabel)
+- #711 Allow building documentation without having rope module installed (@kloczek)
+- #722 Speed up module cache generation by factor of 2 (@tkrabel)
 
 # Release 1.10.0
 
--- a/rope/contrib/autoimport/utils.py
+++ b/rope/contrib/autoimport/utils.py
@@ -20,7 +20,7 @@ def get_package_tuple(
     """
     package_name = package_path.name
     package_type: PackageType
-    if package_name.startswith(".") or package_name == "__pycache__":
+    if package_name.startswith(".") or package_name in ["__pycache__", "site-packages"]:
         return None
     if package_name.endswith((".egg-info", ".dist-info")):
         return None


rope/contrib/autoimport/sqlite.py
```diff
@@ -529,6 +529,7 @@ def _get_packages_from_cache(self) -> List[str]:
             chain(*self._execute(models.Package.objects.select_star()).fetchall())
         )
         existing.append(self.project_package.name)
+        existing.append("site-packages")
```
tkrabel, line 532: Happy to consolidate L531 and L532 into one
```suggestion
        existing += [self.project_package.name, "site-packages"]
```
but actually I like two `append`s better

bagel897:
I should have written logic to handle this already, it may just be missing the regex rules

bagel897:
Here's what I found: https://github.com/python-rope/rope/commit/2182bfe06296a03b953f7cb245452d583cc57875 
I may have assumed all virtual environments began with .venv, You may want to change this logic instead - IE add a check that name is not site-packages. Did we merge the database versioning update? I would purge and re-create all dbs after this change regardless of the approach

bagel897:
Honestly if I was to rewrite this again, I'd abstract the db, parsing and discovery layers out of the main class - it's a bit of a monstrosity. There's also some poor logic in one of the cache generation functions I fixed in a PR I never merged.

tkrabel:
> There's also some poor logic in one of the cache generation functions I fixed in a PR I never merged.

So why not merge it then? :)

lieryan (approved):
I hadn't tested this, but from bird's eye view, this looks reasonable. @bagel897 mentioned that she did a similar fix in an unmerged branch, so if this looks ok to her, it is ok to me as well.

bagel897:
Does this handle cases where venv is in the project root? I am running into that on windows with the autodetect branch. Maybe this is a venv vs virtualenv thing? I think our implementation should be recursive (IE no Lib.site-packages.xyz modules) but I'm not sure how to implement that efficiently

tkrabel-db:
> Does this handle cases where venv is in the project root? I am running into that on windows with the autodetect branch. Maybe this is a venv vs virtualenv thing? I think our implementation should be recursive (IE no Lib.site-packages.xyz modules) but I'm not sure how to implement that efficiently

I think this PR doesn't change the way we treat virtual environments. The PR only doesn't treat `site-packages` like a python package, which makes sense to me. 

That said: @lieryan I saw you approved the PR. Can you merge this? 

@bagel897 if you have any concerns related virtual environments, can we please address this in a separate issue/PR?

bagel897:
Autoimport scans for packages in the python path
image
In this example, we can see both ``../.venv/lib64/python3.12/site-packages`` and also the cwd ``''``
If we detect site-packages, that means a directory called site-packages in the search path - which means a failure to detect venvs. We shouldn't be searching the venv directory from the project root.  The current method is to exclude directories prefixed with a ., which is insufficient.

Here's another failure example:
https://github.com/python-rope/rope/actions/runs/6739504275/job/18321101941?pr=516 
image
Since it searches the venv, it indexes its contents twice - under the ``Lib`` package. We need to exclude all instances of site-packages anywhere in the package or module path. Maybe the solution is to exclude it both at the package and module level?

tkrabel-db:
@bagel897 @lieryan can we discuss this in a follow up issue and merge this PR?
IIUC, my Pr provides an incremental improvement, while not causing any regressions, so I'd not want to delay this improvement unnecessarily.

lieryan:
Hi @tkrabel-db, thanks for creating this PR. I'm currently travelling on vacation, so my availability to look into rope issues and review PRs will be limited until December. 

So don't worry, but I hadn't forgotten about this 🙂, but I won't be able to action the PR or make a new release until December.

tkrabel:
@bagel897 coming back to your comment:

> We need to exclude all instances of site-packages anywhere in the package or module path

I tested my code creating a virtual environment using `venv` and I am fairly confident that my PR does exactly that. It may help looking at my [problem statement](https://github.com/python-rope/rope/issues/722#issue-1972739302) again.

There, you can see we have duplications of the form

```
Name(name=<import_name>, modname=<mod_name>, package=<package>, ...)
Name(name=<import_name>, modname="site-packages."<mod_name>, package='site-packages', ...)
```

Let's use na example to facilitate the conversation.

```
Name(name='BaseName', modname='jedi.api.classes', package='jedi', source=<Source.SITE_PACKAGE: 4>, name_type=<NameType.Class: 7>)
Name(name='BaseName', modname='site-packages.jedi.api.classes', package='site-packages', source=<Source.SITE_PACKAGE: 4>, name_type=<NameType.Class: 7>)
```

Here, you see that we attribute the `BaseName` object to `jedi` in the first entry - the correct match. In the second entry, however, we attribute the same object to a non-existing package we call `site-packages`. 

The change I propose (with your help of pointing me to the correct util; thank you!) captures all cases in which we attribute any object/module to be of root (i.e. the "package") `site-packages`. This means it also captures the failure case you have eluded to earlier, namely `('from Lib.site-packages.pytoolconfig.sources import pytool', 'pytool')`. 

Again, I tested my code and found that it also captures the case you presented.

If you think my PR still has some gaps, pls show them in a reproducible example based on this PR. And thank you for engaging in this PR! It's truly helpful having the original author around :)

lieryan:
Thanks for fixing this issue @tkrabel

PR #751, opened 2024-01-23
Check for ast.Attributes when finding occurrences in f-strings

sandratsy:
# Description

Given the following piece of code:
```python
def xyz(self):
    print(f'{self.abc}')
    return self.abc
```

If I try to Rename attribute/property `abc`, the `abc` in the return statement is correctly renamed, while the `abc` in the f-string is not. This is because `_search_in_f_string()` only looks for ast.Names, while `self.abc` is an ast.Attribute. I have updated `_search_in_f_string()` to look for ast.Attributes as well.

(Ofc, this is not fool-proof, since devs can put whatever they like in f-strings, but I suppose devs are more likely to put Names and Attributes in f-strings than entire function calls.)

The `node.col_offset` for an ast.Attribute points to the start of `self.abc` rather than `abc`, however. To get around this, I've taken the liberty to change `_search_in_f_string()` to return an offset instead of the node. `node.end_col_offset - len(self.name)` will correctly return the index of `abc`.

I used `node.end_col_offset - len(self.name)` instead of `node.col_offset + f_string.index(self.name)` in case the attribute is named `abc.abc`. We want the offset of the 2nd `abc`, not the 1st.


sandratsy:
Hello, any updates on this? Is there anything else I should do?

lieryan:
Thanks for the contribution @sandratsy, this is great. I've added some test cases and type hints but the fix looks sensible.

lieryan:
@all-contributors please add @sandratsy for code

sandratsy:
Thanks v much!

PR #756, opened 2024-02-25
fix: Fix import paths with `site-packages` as part of package name

last-partizan:
# Description

This is another fix, related to #722, i just started using `python-lsp-server` and `rope` for autoimports, and noticed it includes `.direnv.python-3.11.lib.python3.11` as part of module name when importing.

~~We need to cut part of the path to site-packages, to generate correct import name.~~

That was my first attempt to fix this, but turned that wasn't a problem. The problem was, in my case - that local venv `.venv` was included in project resources.

As far as i understand, we have two caches

- `generate_cache` creates project-specific cache
- `generate_module_cache` creates cache of system/venv modules.

And when .venv is in the same dir as a project, it was included in project files.

My current first attempt just filters out path if it has "site-packages" in it, but, maybe we need smarter solution, for example:

`get_python_files` should exclude any file from `get_python_path_folders` (except project root).

For this i need some guidance, @tkrabel @lieryan 

- [ ] I have added tests that prove my fix is effective or that my feature works (not yet)
- [ ] I have updated CHANGELOG.md (not yet)

### DIFF
--- a/rope/contrib/autoimport/parse.py
+++ b/rope/contrib/autoimport/parse.py
@@ -97,6 +97,8 @@ def get_type_object(imported_object) -> NameType:
 
 def get_names(module: ModuleInfo, package: Package) -> List[Name]:
     """Get all names from a module and package."""
+    if "site-packages" in module.modname:
+        return []
     if isinstance(module, ModuleCompiled):
         return list(
             get_names_from_compiled(package.name, package.source, module.underlined)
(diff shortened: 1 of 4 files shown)


rope/base/project.py
```diff
@@ -224,9 +224,22 @@ def __init__(
...
+        extra_ignores = [
+            p.pathlib.relative_to(self.root.pathlib)
+            for p in self.get_python_path_folders()
+            if p.pathlib.is_relative_to(self.root.pathlib)
+            and "site-packages" in p.pathlib.parts
+        ]
+        for p in extra_ignores:
+            # I'm using this approach, because self.prefs.add after _init_prefs
+            # does not work for some reason.
+            prefs["ignored_resources"].append(str(p))
```
last-partizan, line 237: That's my second attempt, and i need feedback here.

Is it good approach, or maybe some other ideas?

PR #788, opened 2024-04-04
Implement preferred_import_style

lieryan:
# Description

Added a new setting (`imports.preferred_import_style`) that controls how rope inserts new import statements.

It can be set to:

- `normal-import` (default) which will insert imports as `import <package>`
- `from-module` which will insert imports as `from <package> import <module>`
- `from-global` which will insert imports as `from <package>.<module> import <object>`

Credit: the solution here is based on [nicoolas25](https://github.com/nicoolas25)'s work on #696, but implements the config settings in a slightly different way.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -69,7 +69,7 @@ autoimport.* Options
 
 .. autopytoolconfigtable:: rope.base.prefs.AutoimportPrefs
 
-import.* Options
+imports.* Options
 ----------------
 
 .. autopytoolconfigtable:: rope.base.prefs.ImportPrefs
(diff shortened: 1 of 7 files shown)


docs/configuration.rst
```diff
@@ -69,6 +69,10 @@ autoimport.* Options
 
 .. autopytoolconfigtable:: rope.base.prefs.AutoimportPrefs
 
+import.* Options
```
lieryan, line 72:
```suggestion
imports.* Options
```

lheckemann:
Just tested this, it works nicely and results in much smaller diffs :) thanks!

nip3o:
@lieryan it would be pretty awesome to be able to use this. Is there something that prevents it from being merged? Anything we could contribute with?

lieryan:
Thanks for checking this PR out, @lheckemann.

@nip3o nothing that I recall, I'll merge this now and include this in a release soon.

PR #795, opened 2024-09-20
Fix mention of Move -> create_move in docs

### LINKED ISSUE #794: `Move` class mentioned in refactoring docs doesn't seem to actually exist
### Description

The [docs on refactorings](https://rope.readthedocs.io/en/latest/library.html#refactorings) mention a `Move` class to be found somewhere in the `rope.refactor` package and suggest the user instantiate it like so:

```python
mover = Move(project, resource, offset)
```

But this class doesn't seem to exist anywhere under `rope.refactor`.

Instead, there are various more specific classes like `MoveMethod` or `MoveModule`, plus a helper function [`create_move`](https://github.com/python-rope/rope/blob/54433b8ca1a6870e68e36ceefe753a71e71eaaaf/rope/refactor/move.py#L33) whose signature seems to match that of the `Move` class currently mentioned in the docs.

So the docs should probably be updated to refer to `create_move` instead.

smheidrich:
# Description

The docs on refactoring mention a `Move` class that doesn't exist in the current codebase. This changes it to `create_move` which does exist and matches the signature of the supposed `Move` class.

Fixes #794

### DIFF
--- a/docs/library.rst
+++ b/docs/library.rst
@@ -523,12 +523,15 @@ Refactorings
 ============
 
 Have a look at ``rope.refactor`` package and its sub-modules.  For
-example for performing a move refactoring you can create a ``Move``
-object like this:
+example for performing a move refactoring you can create an object
+representing this operation (which will be an instance of e.g.
+`MoveMethod`, `MoveModule`, ...) like this:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


lieryan:
LGTM. Thanks for noticing and fixing this, @smheidrich

PR #797, opened 2024-09-20
Add auto-generated API docs

### LINKED ISSUE #796: Automatically generated API docs would be nice
As far as I can tell, there are no automatically generated API docs and users are instead advised to have a look at the source code themselves. E.g. the [docs on refactorings](https://rope.readthedocs.io/en/latest/library.html#refactorings) suggest:

>Have a look at the `rope.refactor` package and its sub-modules.

Needless to say, this is not very convenient.

It would be nice if there were auto-generated API docs for Rope's public interface.

**Describe alternatives you've considered**

- Leave things as they are: People have a needlessly hard time using this library.
- Add more hand-written docs: It's an option but probably much more effort than auto-generation.

**Additional context**
None.

smheidrich:
# Description

Add an auto-generated API reference to Rope's docs.

Fixes #796

## Details (might change as the PR evolves)

- Currently uses `sphinx-autoapi` to generate the docs.
- Lots of warnings during the docs build due to references that can't be found (mostly due to Markdown-like single-backtick inline code in docstrings), but result looks all right.
- Would obviously benefit greatly from enabling [pull request reviews](https://docs.readthedocs.io/en/stable/pull-requests.html) in the ReadTheDocs setting.

### DIFF
--- a/docs/library.rst
+++ b/docs/library.rst
@@ -81,7 +81,7 @@ The first approach uses the ``Project.get_resource()`` method.
 
   myresource = myproject.get_resource('/path/to/resource')
 
-However, it's preferable to use the ``libutils.path_to_resource()``
+However, it's preferable to use the `libutils.path_to_resource()`
 function, because it's more flexible and offers a unified way to create
 resources.  It takes a ``project`` and ``path`` as parameters with an
 optional ``type``.  The ``type`` parameter, with values ``file`` or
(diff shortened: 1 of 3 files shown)


docs/library.rst
```diff
@@ -522,9 +522,10 @@ exception.
 Refactorings
 ============
 
-Have a look at ``rope.refactor`` package and its sub-modules.  For
-example for performing a move refactoring you can create a ``Move``
-object like this:
+Have a look at the `rope.refactor` package and its sub-modules.  For
+example for performing a move refactoring you can create an object
+representing this operation (which will be an instance of e.g.
+`MoveMethod`, `MoveModule`, ...) using `create_move`.
```
smheidrich, line 528: I've made a few adjustments like this one in `library.rst` just to show the benefit of the API docs for readers of the existing docs.

Ideally, this should be done throughout the existing docs, but I'll only do that if I get the green light for the basic idea behind this PR.

Issue #807, opened 2024-11-27
Rope exception when f-string has a # before a variable when using extract method


rlbr:
**Describe the bug**
Rope throws an exception when an f-string has a # before a {variable} when using extract method

For example, this is a valid f-string in python:

```python
f"#{h}"
```

However, selecting this block in an editor and attempting to use code action extract would result in an error being thrown.

**To Reproduce**
1. Code before refactoring:
```python
h = 1
g = f"#{h}"
```

2. Describe the refactoring you want to do

Extract the f-string to a function

4. Expected code after refactoring:

```python
def extracted(h)
    return f"#{h}"
g = extracted(bar)
```

5. Describe the error or unexpected result that you are getting

```
[stderr] 2024-11-27 14:36:45,788 CST - ERROR - pylsp_rope.plugin - Exception when doing workspace/executeCommand: f-string: invalid syntax (<string>, line 2)
[stderr] Traceback (most recent call last):
[stderr]   File "/home/raphael/.venv/testware/lib/python3.9/site-packages/pylsp_rope/plugin.py", line 158, in pylsp_execute_command
[stderr]     return commands[command](workspace, **arguments[0])(
[stderr]   File "/home/raphael/.venv/testware/lib/python3.9/site-packages/pylsp_rope/refactoring.py", line 44, in __call__
[stderr]     rope_changeset = self.get_changes()
[stderr]   File "/home/raphael/.venv/testware/lib/python3.9/site-packages/pylsp_rope/refactoring.py", line 112, in get_changes
[stderr]     rope_changeset = refactoring.get_changes(
[stderr]   File "/home/raphael/.venv/testware/lib/python3.9/site-packages/rope/refactor/extract.py", line 82, in get_changes
[stderr]     new_contents = _ExtractPerformer(info).extract()
[stderr]   File "/home/raphael/.venv/testware/lib/python3.9/site-packages/rope/refactor/extract.py", line 287, in extract
[stderr]     extract_info = self._collect_info()
[stderr]   File "/home/raphael/.venv/testware/lib/python3.9/site-packages/rope/refactor/extract.py", line 314, in _collect_info
[stderr]     self._find_matches(extract_collector)
[stderr]   File "/home/raphael/.venv/testware/lib/python3.9/site-packages/rope/refactor/extract.py", line 328, in _find_matches
[stderr]     for region_match in region_matches:
[...]

PR #808, opened 2024-12-02
Added test with fstring

rlbr:
# Description

Adds a test that reproduces broken behavior with fstrings

Reproduces #807


rlbr:
Fixes #807

rlbr:
All test passed on my machine

Issue #623, opened 2022-12-22
patchedast is currently not handling some `match-case` statement syntaxes


lieryan:
**Is your feature request related to a problem? Please describe.**

#456 adds pattern matching to rope, but it seems like support is incomplete as `patchedast` currently doesn't handle these nodes:

'MatchOr',
'MatchSequence',
'MatchSingleton',
'MatchStar',

This will likely cause extract refactoring to have issues when these syntaxes are being used.

**Describe the solution you'd like**

Implement these nodes.

**Docs on how to implement this**

- https://github.com/python-rope/rope/wiki/Supporting-new-syntax-for-refactoring

PR #819, opened 2025-06-28
supports MatchOr, MatchSequence, MatchStar, MatchSingleton

jheld:
# Description

Please include a summary of the change and which issue is fixed.

Fixes maybe all of #623 

I was initially encountering an issue when trying to implement support for `MatchSingleton`, but I realized that the singleton wasn't actually a `Name`/`NameConstant`, but rather the string representation, at least in the test, so that helped me push through to support it as well.


lieryan:
Thanks for contributing this PR, @jheld. I've merged with a few additional fixes.

@allcontributors please add @jheld for code contribution

PR #824, opened 2025-11-06
Do not use sys.executable for pip uninstall

s-t-e-v-e-n-k:
# Description

The system executable might be an externally managed environment[1], which means pip then refuses to uninstall the installed package, and it bubbles up as a teardown error. Use the venv to avoid that.

1: https://packaging.python.org/en/latest/specifications/externally-managed-environments/


lieryan:
Thank you for investigating and contributing this PR @s-t-e-v-e-n-k.

Closed without merging because this fix is the exact same as #835.

PR #848, opened 2026-07-25
Fix autoimport.sqlite deriving module names from non-package folders

### LINKED ISSUE #823: autoimport.sqlite uses the relative path as module name
**Describe the bug**
There seems to be a difference in the resolution of the module name in autoimport.sqlite and in autoimport.pickle. I don't know if this different logic was intentional when the new implementation was written.

autoimport.sqlite.import_assist returns [('MonitoringDocument', '**shared.src.shared.documents**')]
autoimport.pickle.import_assist returns [('MonitoringDocument', '**shared.documents**')]

The first one is just the relative path and the second one is the right module path.

In the pickle implementation, the function starts in the file's folder and then goes up until it finds a folder that does not contain a __init__.py file -> https://github.com/python-rope/rope/blob/master/rope/base/libutils.py#L106

In the sqlite implementation, the function starts at the root level of the project and goes down to the file. Along the way, all folder names are added to the modname.

**To Reproduce**

1. Clone https://github.com/SimonMonecke/rope-module-name-resolution
2. Run `uv sync`
3. Run `uv run python main.py`

 - Project Python version: 3.13.7
 - Rope Python version: You can use rope as a lib to reproduce the behavior
 - Rope version: 1.14.0
 - Text editor/IDE and version: You can use rope as a lib to reproduce the behavior

**Additional context**
I know that the project structure is a little unusual. Perhaps the solution is that I need to change it and that the sqlite implementation is doing everything right.
Our project is an AWS Lambda function, and another folder contains a project with the code for an ECS task. Both should have access to the code in the shared folder, which is why the shared folder is included as an editable dependency.

deepakganesh78:
Fixes #823

## Root cause

`AutoImport._resource_to_module` in `rope/contrib/autoimport/sqlite.py` derived a module's name from its path *relative to the project root*:

```python
resource_modname = get_modname_from_path(
    resource_path, self.project_package.path, add_package_name=False
)
```

`get_modname_from_path` simply joins every intermediate directory name, so it includes folders that are **not** Python packages. For a layout like

```
<root>/shared/src/shared/__init__.py
<root>/shared/src/shared/documents.py
```

where `shared/` (top) and `src/` are not packages, the sqlite backend reported the module as `shared.src.shared.documents` instead of the correct `shared.documents`.

The pickle backend does not have this bug because it uses `libutils.modname`, which walks *up* the package hierarchy from the resource and stops at the first directory without an `__init__.py`.

## Fix

Use `libutils.modname(resource)` in the sqlite backend's `_resource_to_module`, matching the pickle backend so both agree and intermediate non-package folders are excluded. The now-unused `get_modname_from_path` import is removed.

## Tests

Added `test_resource_to_module_skips_non_package_folders` which builds a `src/mypkg/documents.py` structure (where `src/` is not a package) and asserts the derived modname is `mypkg.documents`. It fails on `master` (`src.mypkg.documents`) and passes with this change.

Full suite: **2092 passed, 8 skipped, 5 xfailed** in a per-repo venv.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # **Upcoming release**
 
+- #823 Fix autoimport.sqlite deriving module names from non-package folders (@deepakganesh78)
 - ...
 
 # Release 1.14.0
--- a/rope/contrib/autoimport/sqlite.py
+++ b/rope/contrib/autoimport/sqlite.py
@@ -44,7 +44,6 @@ from rope.contrib.autoimport.defs import (
 from rope.contrib.autoimport.parse import get_names
 from rope.contrib.autoimport.utils import (
     get_files,
-    get_modname_from_path,
     get_package_tuple,
     sort_and_deduplicate,
     sort_and_deduplicate_tuple,
@@ -654,12 +653,13 @@ class AutoImport:
         assert self.project_package.path
         underlined = underlined if underlined else self.underlined
         resource_path: Path = resource.pathlib
-        # The project doesn't need its name added to the path,
-        # since the standard python file layout accounts for that
-        # so we set add_package_name to False
-        resource_modname: str = get_modname_from_path(
-            resource_path, self.project_package.path, add_package_name=False
-        )
+        # Resolve the module name by walking up the package hierarchy from the
+        # resource (stopping at the first directory without an ``__init__.py``),
+        # rather than deriving it from the resource's path relative to the
+        # project root. The latter incorrectly includes intermediate,
+        # non-package directories in the module name.
+        # See https://github.com/python-rope/rope/issues/823
+        resource_modname: str = libutils.modname(resource)
         return ModuleFile(
             resource_path,
             resource_modname,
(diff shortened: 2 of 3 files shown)


yangfan-yf-yf (changes requested):
The module-name change itself looks correct. The focused selection passes (`5 passed`), and the complete `ropetest/contrib/autoimport/autoimporttest.py` file passes (`17 passed`). I also checked top-level modules, regular packages, a `src/` layout, nested non-package directories, and package `__init__.py`; the sqlite result follows the existing pickle behavior in each case.

I cannot approve the current head yet because it is behind `master` and conflicts in `CHANGELOG.md`. The product and test files merge cleanly; the changelog is the only content conflict.

The exact head also fails the repository pre-commit selection: isort rewrites `rope/contrib/autoimport/sqlite.py` by moving `TYPE_CHECKING` to the sorted position and removing the extra blank line before its block.

Could you update the branch, retain the #823 changelog entry while resolving the conflict, and apply the isort result? I will re-review the resulting head.

PR #859, opened 2026-08-16
Recognise async def method headers in worder

gadievron:
# Description

`worder`'s `is_a_class_or_function_name_in_header` only recognised the `def` and
`class` header prefixes, so an `async def` method header was not treated as a
definition. As a result `find_definition` for an `async def` method called via
`self` resolved to the **call site** instead of the def, and `Rename` from the
call did not reach the def.

```python
class C:
    async def target(self):
        pass
    async def run(self):
        await self.target()   # find_definition here -> the call site, not `def target`
```

Module-level `async def` already resolved correctly (different path); only method
resolution via `self` was affected.

### Fix

Add `"async def"` to the recognised headers and normalise internal whitespace so
`async  def` / `async\tdef` / line-continued forms also match:

```python
return " ".join(prev_word.split()) in ["def", "class", "async def"]
```

The predicate reads the blanked `real_code`, so string/comment text can't spoof
it, and only those three exact strings match — no false-positive surface. The fix
is version-independent (`async def` has been valid since Python 3.5).

Tests cover `find_definition` on an async method (canonical and tab/extra-space
spellings), an end-to-end `Rename`, and two regression controls (sync methods and
module-level async defs still resolve).

---

# Checklist

- [x] I have added tests that prove my fix is effective

### DIFF
--- a/ropetest/contrib/findittest.py
+++ b/ropetest/contrib/findittest.py
@@ -3,6 +3,7 @@ from textwrap import dedent
 
 from rope.base import exceptions
 from rope.contrib.findit import find_definition, find_implementations, find_occurrences
+from rope.refactor.rename import Rename
 from ropetest import testutils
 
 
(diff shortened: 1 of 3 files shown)


yangfan-yf-yf (approved):
I reviewed `93082f7` and found the predicate change appropriately narrow. I verified canonical `async def`, double-space, tab-separated, and backslash-continued method headers: current `master` resolves each `self.target()` reference to the call offset, while this head resolves all four to the method definition offset. Existing synchronous methods and module-level async functions remain covered.

The full find-definition file passes (`17 passed`), and the code-assist plus rename suites pass (`267 passed`). `git diff --check` also passes. I found no overlapping open implementation and no merge blocker. LGTM.

ropetest/contrib/findittest.py
```diff
@@ -145,3 +146,109 @@ def test_find_definition_in_other_modules(self):
         result = find_definition(self.project, code, code.index("var"))
         self.assertEqual(mod1, result.resource)
         self.assertEqual(0, result.offset)
+
+    # worder's is_a_class_or_function_name_in_header only recognised the
+    # canonical `def`/`class` prefixes, so an `async def` method header was
+    # not treated as a definition and find_definition resolved a call via
+    # self to the call site itself (never the def).
+
```
lieryan: Remove

ropetest/contrib/findittest.py
```diff
@@ -145,3 +146,109 @@ def test_find_definition_in_other_modules(self):
...
+        call_offset = code.rindex("target")
+        changes = Rename(self.project, mod, call_offset).get_changes("renamed")
+        self.project.do(changes)
+        result = mod.read()
+        # both the call and the def must be renamed -- not silently orphaned
+        self.assertEqual(0, result.count("target"))
+        self.assertEqual(2, result.count("renamed"))
+
+    def test_find_definition_of_sync_method_still_resolves(self):
+        # regression control: sync methods must keep resolving
```
lieryan: Regression test already exist, need to integrate better with existing test suite

ropetest/contrib/findittest.py
```diff
@@ -145,3 +146,109 @@ def test_find_definition_in_other_modules(self):
...
+        # `async\tdef` (tab-separated) is valid Python; the normalised
+        # predicate must recognise it too.
+        code = dedent("""\
+            class C(object):
+                async def target(self):
+                    pass
+
+                async def run(self):
+                    await self.target()
+        """).replace("async def", "async\tdef")
```
lieryan: just use the literals in the triple quoted string?

ropetest/contrib/findittest.py
```diff
@@ -145,3 +146,109 @@ def test_find_definition_in_other_modules(self):
...
+            class C(object):
+                async def target(self):
+                    pass
+
+                async def run(self):
+                    await self.target()
+        """)
+        mod.write(code)
+        call_offset = code.rindex("target")
+        changes = Rename(self.project, mod, call_offset).get_changes("renamed")
```
lieryan: rename test belongs in renametest.py

PR #863, opened 2026-08-20
Handle parentheses around MatchSequence more correctly

lieryan:
# Description

Handle various corner cases around MatchSequence. It's complicated by the fact that the MatchSequence syntax allows multiple kinds of parentheses and even no parentheses and the ast doesn't give us any clue which is being used.

### DIFF
--- a/rope/refactor/patchedast.py
+++ b/rope/refactor/patchedast.py
@@ -805,33 +805,41 @@ class _PatchingASTWalker:
 
     def _MatchSequence(self, node):
         if node.patterns:
-            opening_paren = self.lines[
-                (node.lineno, node.col_offset) : (
-                    node.patterns[0].lineno,
-                    node.patterns[0].col_offset,
-                )
-            ].strip()
-            closing_paren = self.lines[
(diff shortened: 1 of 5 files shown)


rope/refactor/patchedast.py
```diff
@@ -804,7 +804,33 @@ def _match_case(self, node):
...
+                    node.patterns[0].lineno,
+                    node.patterns[0].col_offset,
+                )
+            ]
+            closing_paren = self.lines[
+                (node.patterns[-1].end_lineno, node.patterns[-1].end_col_offset) : (
+                    node.end_lineno,
+                    node.end_col_offset,
+                )
+            ]
```
lieryan: Fix formatting of these lines

rope/refactor/patchedast.py
```diff
@@ -804,7 +804,33 @@ def _match_case(self, node):
...
+                *closing_paren,
+            ]
+        else:
+            children = [
+                self.lines[
+                    (node.lineno, node.col_offset) : (
+                        node.end_lineno,
+                        node.end_col_offset,
+                    )
+                ]
```
lieryan: And these too

PR #865, opened 2026-08-26
Recognize match pattern captures as assignments

### LINKED ISSUE #861: Rename and inline refactor did not recognize match-cases statements as variable assignment
**Describe the bug**

The match statement allows pattern capture which effectively acts like an assignment statement. Currently they are not recognised as assignment statement that can define a variable.

**To Reproduce**
1. Code before refactoring:

```python
x = [1, 2, 3]
match x:
    case [1, *_] as old_name_1:
        print(old_name_1)
    case [2, *old_name_2]:
        print(old_name_2)
```

2. Describe the refactoring you want to do: rename `old_name_xxx` to `new_name_xxx`

```python
x = [1, 2, 3]
match x:
    case [1, *_] as new_name_1:
        print(new_name_1)
    case [2, *new_name_2]:
        print(new_name_2)
```

```
RefactoringError: Rename refactoring should be performed on resolvable python identifiers
```

yangfan-yf-yf:
# Description

Pattern capture names on `MatchAs`, `MatchStar`, and `MatchMapping` are string fields, so the scope visitor did not register them as assignments. Rename consequently rejected those names as unresolved. Inline could also replace a capture with a value pattern, changing program semantics.

This registers the three capture forms in their enclosing scope, returns a conservative unknown inference result for pattern assignments, and rejects inline whenever the selected name has a pattern-capture assignment. Rename now updates both capture definitions and references.

Fixes #861

# Checklist

- [x] I have added tests that prove the fix is effective
- [x] I have updated `CHANGELOG.md`

# Validation

- Full suite: 2121 passed, 12 skipped, 5 xfailed
- Relevant selection: 442 passed, 10 skipped
- All configured pre-commit checks for the changed Python files pass
- `git diff origin/master...HEAD --check` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - #853 Implement patchedast handlers TypeVar
 - #847 Avoid printing autoimport syntax errors (@yangfan-yf-yf)
 - #623, #819, #863 Support MatchOr, MatchSequence, MatchStar (@jheld, @lieryan)
+- #861 Recognize pattern captures for rename and reject unsupported inline operations (@yangfan-yf-yf)
 
 # Release 1.14.0
 
--- a/rope/base/oi/soi.py
+++ b/rope/base/oi/soi.py
@@ -6,7 +6,7 @@ package.
 """
 
 import rope.base.builtins  # Use full qualification for clarity.
-from rope.base import arguments, evaluate, pynames, pyobjects, utils
+from rope.base import arguments, ast, evaluate, pynames, pyobjects, utils
 from rope.base.oi.type_hinting.factory import get_type_hinting_factory
 
 _ignore_inferred = utils.ignore_exception(pyobjects.IsBeingInferredError)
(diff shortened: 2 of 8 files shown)


yangfan-yf-yf:
I merged the current master into 48c33d9 to resolve the conflict. Only the changelog entries conflicted; both the pattern-capture entry and the new `is_dir()` entry are retained. The pattern-capture implementation is unchanged by the merge.

Validation on Windows / Python 3.12:

- Full suite: 2121 passed, 12 skipped, 5 xfailed, 1 failed.
- The failure is `ropetest/contrib/autoimporttest.py::test_search_submodule`. It also fails in a separate checkout of unmodified master at f005ac7 with the same assertion: searching for `mod1` from the installed `external_fixturepkg` returns no matches.
- All configured pre-commit hooks for the PR's Python files pass, and the diff has no whitespace errors.

The branch update is pushed for the upstream checks to run.

PR #874, opened 2026-09-13
Treat type alias statements as name assignments

### LINKED ISSUE #862: Rename and inline refactor did not recognize type alias statements as variable assignment
**Describe the bug**

The type alias statement is a new kind of assignment statement since Python 3.12+. Currently they are not recognised as assignment statement that can define a variable.

**To Reproduce**
1. Code before refactoring:

```python
type old_name_1 = int
x: old_name_1 = 1

type Alias[old_name_2] = list[old_name_2]
print(old_name_2) # this should be untouched
```

2. Describe the refactoring you want to do: rename `old_name_xxx` to `new_name_xxx`

```python
type new_name_1 = int
x: new_name_1 = 1

type Alias[new_name_2] = list[new_name_2]
print(old_name_2) # this should be untouched
```

```
RefactoringError: Rename refactoring should be performed on resolvable python identifiers
```

TanbirRamim:
# Description

Rename and inline refactoring failed on `type X = ...` statements because the scope visitor never registered the alias name, so rope could not resolve it. The scope visitor now treats a type alias like an assignment and records the alias name. A generic alias like `type A[T] = list[T]` can still be renamed, but it gets no assigned value, so inlining it is refused instead of producing `list[T][int]`. This does not cover renaming the type parameter itself (`old_name_2` in the issue), which needs its own scope for the alias and is a bigger change. I added tests for rename and inline that only run on Python 3.12 and newer, and the whole ropetest suite passes on 3.12.

Fixes #862


yangfan-yf-yf (changes requested):
I tested `741e993bed633088ccd687985b85f3d549e77775` on Windows with Python 3.12.3. The six added tests pass, and the complete rename/inline test files pass (205 tests, including those six). Ordinary and generic alias renaming also work in my additional checks, including renaming from an import in another module. Generic alias inlining is refused as described.

I found a correctness issue in the newly enabled **non-generic alias inlining**: recording `node.value` as an ordinary assigned value loses the alias object's runtime behavior and its lazy evaluation. Two valid inputs become programs that raise exceptions after inlining.

Python documents both the distinct alias object and lazy `__value__` evaluation in [TypeAliasType](https://docs.python.org/3.12/library/typing.html#typing.TypeAliasType).

For example, inlining `Alias` in:

```python
type Alias = int
value = Alias.__value__
```

produces:

```python
value = int.__value__
```

The original evaluates to `int`; the result raises `AttributeError: type object 'int' has no attribute '__value__'`.

There is also a failure even when the alias is used only in an annotation:

```python
type Alias = Later
def f(x: Alias):
    pass
class Later:
    pass
```

Inlining removes the alias and changes the annotation to `x: Later`, making module execution fail with `NameError: name 'Later' is not defined`. The original module executes successfully because the alias value is lazy.

I ran both examples against the PR and its base, `f005ac786da7d556fdf5733e9b9ddae1eaf27242`. The base refuses the inline operation and leaves the working code unchanged; this head accepts it and produces the failures above.

Please preserve these semantics or reject the affected inline operations before returning edits, with regression tests for both cases. Keeping the new alias name resolution while refusing alias inlining until these semantics are handled would also be a safe narrower change. This concern is about the non-generic inline behavior added here, not the explicitly deferred type-parameter renaming support.

TanbirRamim:
Thanks @yangfan-yf-yf, good catches. I merged master (resolved the CHANGELOG conflict) and took the narrower route you suggested: type alias names are still resolved for rename, but the alias value is no longer recorded as an assignment, so inlining any type alias is now refused. Your two examples are added as regression tests (`Alias.__value__` and the lazily evaluated forward reference). Full `ropetest` suite passes locally on Python 3.13.

